PR #1, opened 2021-07-11
Photostructure

mcmikemn:
I'm opening this PR as practice. I don't think you should merge the photostructure branch into master, but the current version works. Photostructure itself does not work with S3, so I separated the S3 instructions out in the readme - those instructions work as far as setting everything up for Photostructure with an s3fs mount, it's just that Photostructure doesn't seem to function with it (possibly s3fs is just too slow).

Let me know if you have any comments or suggestions.


EnigmaCurry:
Closing this, but you can re-open it if you work on it again.

PR #2, opened 2021-07-25
Split ACME certificate resolvers based on environment.

EnigmaCurry:
The Traefik config only has one certificate resolver right now: `default`, and you choose between Let's Encrypt staging or production environments by setting the `ACME_CA_SERVER` variable in `traefik/.env`. This is a global setting and affects all deployments.

This PR splits ACME into two files. It removes the `default` certresolver and creates two new ones: `staging` and `production`, and so each deployment can choose which environment they would like their certificate created in. Traefik will store TWO acme.json files:

 * For production: `/data/acme-production.json` 
 * For development/staging: `/data/acme-staging.json`

If you redeploy Traefik, in the `traefik` directory, with `docker-compose up -d`, all of your existing deployments will now have invalid certificates. You can check the traefik logs to see why:

```
time="2021-07-25T21:40:37Z" level=error msg="the router whoami@docker uses a non-existent resolver: default"
time="2021-07-25T21:40:37Z" level=error msg="the router piwigo@docker uses a non-existent resolver: default"
```

(On this server I only have deployments for `whoami` and `piwigo`.)

This is because all of the existing deployments use the `default` certificate resolver, which has now been removed. You need to add the `ACME_CERT_RESOLVER` variable to all of your deployment `.env` files and choose `staging` or `production` depending on whether you need a production certificate or not, and redeploy (If you do not add the variable, the service will have an invalid `TRAEFIK DEFAULT CERT` certificate.)

All of the example `.env-dist` files now have added a `ACME_CERT_RESOLVER` variable with the default for the `staging` environment. You need to copy these to any existing `.env` files.

If you want to retain all of your old production certificates, you will need to copy your old `acme.json` to the new `acme-production.json` file:

Enter into the shell of the container running Traefik:

```
docker exec -it traefik /bin/sh
```

Now inside the traefik container, run:

```
awk '/default/ && !done { gsub(/default/, "production"); done=1}; 1' /data/acme.json > /data/acme-production.json
chmod 600 /data/acme-production.json
```

(This will copy acme.json to acme-production.json and automatically edit the name inside from "default" to "production".)

Press `Ctrl-D` to exit the container.

Restart traefik again. In the `traefik` directory run `docker-compose up -d`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,16 +87,15 @@ HTTP reverse proxy and load balancer.
 
 Copy `.env-dist` to `.env` and edit the following:
 
- * `ACME_CA_SERVER` this is the Let's Encrypt API (ACME) server to use. 
- 
-   * For development/staging use
-     `https://acme-staging-v02.api.letsencrypt.org/directory`
-   * For production use `https://acme-v02.api.letsencrypt.org/directory`
-   
  * `ACME_CA_EMAIL` this is YOUR email address, where you will receive notices
    from Let's Encrypt regarding your domains and related certificates.
 
 To start Traefik, go into the traefik directory and run `docker-compose up -d`
+
+All other services defines the `ACME_CERT_RESOLVER` variable in their respective
+`.env` file. There are two different environments defined, `staging` (default)
+and `production`. Staging will create untrusted TLS certificates for testing
+environments. Production will generate browser trustable TLS certificates.
 
 ## Gitea
 
--- a/baikal/.env-dist
+++ b/baikal/.env-dist
@@ -1,1 +1,3 @@
 BAIKAL_TRAEFIK_HOST=cal.example.com
+# Choose Let's Encrypt 'staging' or 'production' environment:
+ACME_CERT_RESOLVER=staging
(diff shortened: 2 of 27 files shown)


EnigmaCurry:
So as an update... this is not entirely broken, but its kind of broken. But if you know how it works, you can still use it and you can just avoid the problems it has.

See https://github.com/traefik/traefik/issues/8311

You cannot freely switch between staging and production certresolvers anytime you want. For each domain Host rule, you have to pick the environment and stick with it. So that means you can use a given domain `whoami.example.com` with a staging LE certificate, but that domain has to stay on a staging LE certificate from now on (unless you want to delete or edit the acme-staging.json to remove the certificate(s) manually). If you want to use a production certificate, it has to be for a different domain name. So this just means that if you want to switch, you have to change the certresolver *and* the the Host router rule for the given domain name. For example, the whoami container, for development environment should use the router rule `Host(whoami.test.example.com)` and the `staging` ACME_CERT_RESOLVER, while the production environment should use a different router rule `Host(whoami.example.com)` and the `production` ACME_CERT_RESOLVER.

PR #35, opened 2023-02-16
Add Plausible Analytics

jessopb:
Adds plausible analytics package.


plausible/docker-compose.yaml
```diff
@@ -0,0 +1,54 @@
...
+    restart: always
+
+  plausible_db:
+    # supported versions are 12, 13, and 14
+    image: postgres:14-alpine
+    restart: always
+    volumes:
+      - db-data:/var/lib/postgresql/data
+    environment:
+      - POSTGRES_PASSWORD=postgres
```
jessopb, line 14: Gotta config these

plausible/docker-compose.yaml
```diff
@@ -0,0 +1,54 @@
...
+    #   - 8000:8000
+    environment:
+      - PLAUSIBLE_TRAEFIK_HOST
+      - BASE_URL=https://${PLAUSIBLE_TRAEFIK_HOST}
+      - SECRET_KEY_BASE=${PLAUSIBLE_SECRET_KEY_BASE}
+    labels:
+      - "traefik.enable=true"
+      - "traefik.http.routers.plausible.rule=Host(`${PLAUSIBLE_TRAEFIK_HOST}`)"
+      - "traefik.http.routers.plausible.entrypoints=websecure"
+      - "traefik.http.services.plausible.loadbalancer.server.port=8000"
```
EnigmaCurry, line 46: Treafik will figure this out from the image EXPOSE without needing to specify server.port

plausible/docker-compose.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - ./clickhouse/clickhouse-user-config.xml:/etc/clickhouse-server/users.d/logging.xml:ro
+    ulimits:
+      nofile:
+        soft: 262144
+        hard: 262144
+
+  plausible:
+    image: plausible/analytics:latest
+    restart: always
+    command: sh -c "sleep 10 && /entrypoint.sh db createdb && /entrypoint.sh db migrate && /entrypoint.sh run"
```
EnigmaCurry, line 31: this is weird to call entrypoint, because command is used as an argument *to* the entrypoint. but maybe its correct if the entrypoint is a no-op when you pass an arg like this? 

Is the sleep necessary? I haven't needed to wait for the database to come up before, as long as I use depends_on (which IIRC it uses a healthcheck to determine that db is up or not)

plausible/docker-compose.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - "traefik.http.routers.plausible.entrypoints=websecure"
+      - "traefik.http.services.plausible.loadbalancer.server.port=8000"
+
+volumes:
+  db-data:
+    driver: local
+  event-data:
+    driver: local
+  geoip:
+    driver: local
```
EnigmaCurry, line 54: I think it should use whatever driver the docker engine is configured to use, so please remove the drivers.

plausible/docker-compose.yaml
```diff
@@ -0,0 +1,54 @@
...
+    environment:
+      - POSTGRES_PASSWORD=postgres
+
+  plausible_events_db:
+    image: clickhouse/clickhouse-server:22.6-alpine
+    restart: always
+    volumes:
+      - event-data:/var/lib/clickhouse
+      - ./clickhouse/clickhouse-config.xml:/etc/clickhouse-server/config.d/logging.xml:ro
+      - ./clickhouse/clickhouse-user-config.xml:/etc/clickhouse-server/users.d/logging.xml:ro
```
EnigmaCurry, line 22: Move these bind mounts to a named volume.

plausible/docker-compose.yaml
```diff
@@ -0,0 +1,54 @@
+version: "3.3"
```
EnigmaCurry, line 1: 3.9 is what the other files have.

plausible/docker-compose.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - event-data:/var/lib/clickhouse
+      - ./clickhouse/clickhouse-config.xml:/etc/clickhouse-server/config.d/logging.xml:ro
+      - ./clickhouse/clickhouse-user-config.xml:/etc/clickhouse-server/users.d/logging.xml:ro
+    ulimits:
+      nofile:
+        soft: 262144
+        hard: 262144
+
+  plausible:
+    image: plausible/analytics:latest
```
EnigmaCurry, line 29: this should probably not use `latest`. I wouldcreate an env var called PLAUSIBLE_VERSION,  find the latest tagged release and set the default, and then reference it here. that way the user can pin the version and they know they can upgrade it.

plausible/docker-compose.yaml
jessopb: Copied files in Dockerfile build

EnigmaCurry (approved):

jessopb:
Changed the service names back to their originals. I recall that trying to change them may require more troubleshooting.

EnigmaCurry:
> Changed the service names back to their originals. I recall that trying to change them may require more troubleshooting.

yea service names are also the DNS names, so if those containers have hard coded those names then the resolve fails. working service is better than redundant naming :+1:

EnigmaCurry:
Not a blocker, but I get this error on arm64:

```
 ⠋ mail The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested  0.0s
```

events db log has this warning several times:

```
plausible-plausible_events_db-1  | 2023.02.19 23:06:40.802432 [ 43 ] {} <Warning> Application: Listen [::]:8123 failed: Poco::Exception. Code: 1000, e.code() = 0, DNS error: EAI: Address family for hostname not supported (version 22.6.9.11 (official build)). If it is an IPv6 or IPv4 address and your host has disabled IPv6 or IPv4, then consider to specify not disabled IPv4 or IPv6 address to listen in <listen_host> element of configuration file. Example for disabled IPv6: <listen_host>0.0.0.0</listen_host> . Example for disabled IPv4: <listen_host>::</listen_host>
```

The docker network only has ipv4.

EnigmaCurry:
I think we need to figure out how to disable public registration. Maybe require HTTP basic auth for all routes except those required for the tracking stuff?

EnigmaCurry:
I found how you are supposed to be able to disable registration, but it doesn't seem to be working:

https://github.com/plausible/analytics/blob/8f85b110aa43c17694a103df81fdf9dc098f6c19/config/runtime.exs#L170-L173

EnigmaCurry (approved):

PR #46, opened 2023-02-28
Externalize Postgres

### LINKED ISSUE #45: Services should have the option to externalize PostgreSQL
Any service that uses postgres should be able to deploy with an instance in the same docker-compose.yaml (embedded postgres) or to use an external (or third part) DB / DBaaS.

Build common tooling into `make config` to ask questions about the server, username, keys, passwords, etc.

EnigmaCurry:
This will fix #45  and add the option to any service that requires postgresql to define an external database.


EnigmaCurry:
Working on pgbackrest first and will fix #46 after that.

PR #51, opened 2023-04-09
Audiobookshelf

mcmikemn:
I think audiobookshelf is potentially useful. I don't really understand d.rymcg instances so I just deleted the instance.yaml that `d.rymch.tech create` created. And there's possibly other ways to improve this before merging into master.

Also, I apparently had a bunch of commits I forgot about when I pushed to this branch. What's the best way to fix that?


EnigmaCurry:
> Also, I apparently had a bunch of commits I forgot about when I pushed to this branch. What's the best way to fix that?

You can just push more commits, when this gets merged they will all be "squashed" into one new commit. I'll mark this as a draft PR for now.

mcmikemn:
> > Also, I apparently had a bunch of commits I forgot about when I pushed to this branch. What's the best way to fix that?
> 
> You can just push more commits, when this gets merged they will all be "squashed" into one new commit. I'll mark this as a draft PR for now.

I don't have any more commits. I think it's ready to go, unless I didn't handle instances well or there's anything else you think should be improved.

EnigmaCurry:
OK, I'll check it out today. I notice theres some remnants of an XBS config here [xbs/api/settings.json](https://github.com/EnigmaCurry/d.rymcg.tech/pull/51/files#diff-23b78ac9d3485abcfc89d8874ae40b269170bcbed9b6a8afae630131843b6040)

mcmikemn:
Yes, those are some of my past commits that I had forgotten about and mistakenly pushed into this branch.

EnigmaCurry:
> Yes, those are some of my past commits that I had forgotten about and mistakenly pushed into this branch.

ok so if you make a new commit that deletes those, it will all get squashed so its not in the final merge.

mcmikemn:
I think that's it. There are some merge->master commits, but it doesn't look like those need un-doing.

audiobookshelf/docker-compose.yaml
```diff
@@ -0,0 +1,23 @@
...
+  audiobooks:
+  podcasts:
+  config:
+  metadata:
+    
+services:
+  audiobookshelf:
+    image: ghcr.io/advplyr/audiobookshelf:latest
+    ports:
+      - 13378:80
```
EnigmaCurry, line 12: never specify a port when you are using traefik. Traefik sets the port, not the service being proxied.

audiobookshelf/docker-compose.yaml
```diff
@@ -0,0 +1,23 @@
+version: "3.7"
+volumes:
+  audiobooks:
+  podcasts:
+  config:
+  metadata:
+    
+services:
+  audiobookshelf:
+    image: ghcr.io/advplyr/audiobookshelf:latest
```
EnigmaCurry, line 10: we shouldn't use `latest`, but instead it should have an AUDIOBOOKSHELF_VERSION variable in the config.

audiobookshelf/.env-dist
```diff
@@ -0,0 +1,25 @@
...
+# HTTP Basic Authentication
+# Use `make config` to generate and configure this setting interactively.
+# Set multiple htpasswd encoded username/passwords comma separated; blank to disable
+# See: https://doc.traefik.io/traefik/middlewares/http/basicauth/
+AUDIOBOOKSHELF_HTTP_AUTH=
+
+# Development mode
+## By default, this flag does nothing. But you can use it in docker-compose.instance.yaml however you want.
+## For example you may wish to detect development mode and optionally turn on a live-reload mode of your service.
+AUDIOBOOKSHELF_DEVELOPMENT_MODE=false
```
EnigmaCurry, line 25: we should disable http authentication, and dev mode doesn't make sense here.

EnigmaCurry:
I added a review, and then fixed it myself. I added the instance template back in. Even though audiobookshelf supports multiple users, it still contains universal configuration and a 'root' user. So multiple instances would allow, eg. mike and ryan to each own a separate instances with different configs and volumes, but on a shared droplet, where each have separate root users/backups.

PR #55, opened 2023-04-11
added mqtt config to traefik.yml

mcmikemn:
you had me add these mqtt configs to traeik.yml a while ago - i commited but never pushed. not that commit is gone and I'm trying to do it right. However, please double-check the configs (especially the white space, as I think yaml cares about white space).

### DIFF
--- a/traefik/config/traefik.yml
+++ b/traefik/config/traefik.yml
@@ -111,6 +111,19 @@ entryPoints:
   #@ if data.values.mqtt_entrypoint_enabled == "true":
   mqtt:
     address: #@ data.values.mqtt_entrypoint_host + ":" + data.values.mqtt_entrypoint_port
+    http:
+      tls:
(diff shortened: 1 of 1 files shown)


EnigmaCurry:
I realize I'm the one who suggested doing this, but the Traefik documentation doesn't say this is a valid config, they only show its valid for http, not tcp, and mosquitto isn't using http. (https://doc.traefik.io/traefik/routing/entrypoints/#tls)

I tested without this PR to see if there is an issue, but it worked fine for me, using default install options and no DNS challenge.

I did not run traefik `make certs`, and so it automatically created a new certificate for the mqtt domain only. 

tested with `openssl s_client -tls1_2 -connect mqtt.mtmp.rymcg.tech:8883`

So I think if you remove this from your config, it should still work, even on a fresh droplet.

EnigmaCurry:
closing, if you can reproduce the problem, please open another issue.

PR #64, opened 2023-04-29
added grocy

mcmikemn:
Surely this needs some cleanup, but it works.


mcmikemn:
I don't think I'm going to use grocy for my grocery _or_ chore needs. It's working, though, so you can either add it to d.ry or just delete it.

EnigmaCurry:
still planning on checking it out, thanks!

EnigmaCurry:
I rebased this on the simplified docker-compose.instance.yaml that comes from the whoami example (the one you used from `d.rymcg.tech create` is inferior and still needs work.) I added some of the extra env vars to docker-compose.yaml otherwise they were not seen.

PR #75, opened 2023-05-09
Wordpress

EnigmaCurry:
Merging this from https://github.com/jessopb/wp_d_rymcg


mcmikemn:
This works for me.

I noticed that WordPress 6.2.2 was installed for me even though my `.env_` file used the default value of WP_VERSION from `.env_dist`: `WP_VERSION=6.1.1`. I think this is because 6.1.1 is no longer a supported tag for the image, so the tag was ignored and the main image was used. I don't think this needs any fixing in d.ry, as it's a user-entered variable anyway.

EnigmaCurry:
I've rebased and made some config tweaks, looks good to me!

@jessopb can you please review/approve? Thank you!

PR #80, opened 2023-07-21
Qw configs

mcmikemn:
added more qbittorrent configs to .yaml .conf. I think it would be good to just include all possible configs in both files and .env-dist so the only thing for the d.ry user to do is set their values, but I can't find a complete list of configs. Many of the configs dont exist in the .conf file of the live install unless you change the value in the UI to something other than default, and there's maybe 100-200 configs? So I'm just saving the configs I use or might want to use.


EnigmaCurry:
I agree it would be nice to have them all in .env-dist, but grepping the code it doesnt appear there is an easy to use list anywhere. I've added the new ones you made to .env-dist

mcmikemn:
I changed the default to "false" for QBITTORRENT_SessionUseAlternativeGlobalSpeedLimit and QBITTORRENT_SessionBandwidthSchedulerEnabled so that the other three variables will be ignored and don't need a default value. qBittorrent code might have assumed "false" as defaults for these 2 vars anyway, but I don't know.

I also moved 3 of the vars to a different category to match the category that qBittorrent saves them in.

EnigmaCurry (approved):

EnigmaCurry (dismissed):

EnigmaCurry (approved):
LGTM Thanks @mcmikemn

PR #83, opened 2023-07-21
Photoprism

mcmikemn:
Added Photoprism. 

You can configure your `.env` to pull a "Community Edition" container that's under a [GNU Affero General Public License](https://www.photoprism.app/oss/license), or you can pull an "extended functionality" container that's under their [Plus License](https://www.photoprism.app/plus/license). Very brief description [here](https://docs.photoprism.app/getting-started/docker-compose/#photoprism-plus).

A weird thing is that you can self-host this for free, but if you want the extended features then you need to pay them (see ["Compare Feaures"](https://www.photoprism.app/features)). I don't know how this could work unless they sent you some kind of code you plugged into your Photoprism instance, which doesn't seem to be the case, or unless your self-hosted instance of Photoprism communicated with their servers for authorization, which seems yucky.


photoprism/.env-dist
```diff
@@ -0,0 +1,108 @@
...
+## Traefik IP whitelist filter to block access to the client interface:
+## (does not affect peer connections)
+PHOTOPRISM_IP_SOURCERANGE=0.0.0.0/0
+
+
+PHOTOPRISM_ADMIN_USER=admin
+# initial admin password (8-72 characters)
+PHOTOPRISM_ADMIN_PASSWORD=password
+# server URL in the format "http(s)://domain.name(:port)/(path)"
+PHOTOPRISM_SITE_URL=http://localhost:2342/
```
EnigmaCurry, line 35: this default URL seems wrong

photoprism/docker-compose.yaml
```diff
@@ -0,0 +1,132 @@
...
+services:
+  photoprism:
+    image: photoprism/photoprism:${PHOTOPRISM_VERSION}
+    ## If the service gets stuck in a restart loop, this points to a memory, filesystem, network, or database issue:
+    ## https://docs.photoprism.app/getting-started/troubleshooting/#fatal-server-errors
+    restart: unless-stopped
+    stop_grace_period: 10s
+    depends_on:
+      - mariadb
+# are these security options safe? or do i want `no-new-privileges:true` and `cap_drop:      - ALL`?
```
EnigmaCurry, line 36: no-new-privileges:true is always a good default, as long as you test that it is working, it will mean it cannot get more privilges than you've defined here. cap_drop: all is a more drastic approach that drops all privileges and you need to add every single one it needs; this usually breaks the app, but you just test it adding more capapilities until it works. 

I'm guessing you copied these from the photoprism config, which is fine, you might try using the no-new-privilges at least though.

photoprism/.env-dist
```diff
@@ -0,0 +1,108 @@
...
+# FFmpeg encoding bitrate limit in Mbit/s (default: 50)
+# PHOTOPRISM_FFMPEG_BITRATE=32
+# FFmpeg encoding resolution limit in pixel height (default: 4096)
+# PHOTOPRISM_FFMPEG_RESOLUTION=1920
+## Run as a non-root user after initialization (supported: 0, 33, 50-99, 500-600, and 900-1200):
+PHOTOPRISM_UID=1000
+PHOTOPRISM_GID=1000
+PHOTOPRISM_UMASK=0000
+# MariaDB or MySQL database user password - this should match MARIADB_PASSWORD
+PHOTOPRISM_DATABASE_PASSWORD=insecure
```
EnigmaCurry, line 53: passwords should be randomized in the Makefile

photoprism/Makefile
```diff
@@ -0,0 +1,44 @@
...
+	@${BIN}/reconfigure_ask ${ENV_FILE} PHOTOPRISM_TRAEFIK_HOST "Enter the photoprism domain name" photoprism${INSTANCE_URL_SUFFIX}.${ROOT_DOMAIN}
+	@${BIN}/reconfigure ${ENV_FILE} PHOTOPRISM_INSTANCE=$${instance:-default}
+	@${BIN}/reconfigure_htpasswd ${ENV_FILE} PHOTOPRISM_HTTP_AUTH default=no
+	
+	@echo ""
+	@${BIN}/reconfigure_ask ${ENV_FILE} PHOTOPRISM_ADMIN_USER "Enter the admin user name" ${PHOTOPRISM_ADMIN_USER}
+	@echo ""
+	@${BIN}/reconfigure_ask ${ENV_FILE} PHOTOPRISM_ADMIN_PASSWORD "Enter the admin password" ${PHOTOPRISM_ADMIN_PASSWORD}
+	@echo ""
+	@${BIN}/reconfigure_ask ${ENV_FILE} PHOTOPRISM_ORIGINALS_VOLUME "Enter the admin password" ${PHOTOPRISM_ORIGINALS_VOLUME}
```
EnigmaCurry, line 20: Wrong text for question

photoprism/.env-dist
```diff
@@ -0,0 +1,108 @@
...
+# this should match PHOTOPRISM_DATABASE_PASSWORD
+MARIADB_PASSWORD=insecure
+MARIADB_ROOT_PASSWORD=insecure
```
EnigmaCurry, line 56: Why three environment variables for db password? These should just reuse PHOTOPRISM_DATABASE_PASSWORD in docker-compose.yaml and not exposed here.

photoprism/.env-dist
```diff
@@ -0,0 +1,108 @@
...
+# MariaDB or MySQL database server (hostname:port)
+PHOTOPRISM_DATABASE_SERVER=mariadb:3306
+# MariaDB or MySQL database schema name
+PHOTOPRISM_DATABASE_NAME=photoprism
+# MariaDB or MySQL database user name
+PHOTOPRISM_DATABASE_USER=photoprism
+MARIADB_AUTO_UPGRADE=1
+MARIADB_INITDB_SKIP_TZINFO=1
+MARIADB_DATABASE=photoprism
+MARIADB_USER=photoprism
```
EnigmaCurry, line 108: I see no reason to expose any of these variables

photoprism/.env-dist
```diff
@@ -0,0 +1,108 @@
...
+PHOTOPRISM_DATABASE_PASSWORD=insecure
+# this should match PHOTOPRISM_DATABASE_PASSWORD
+MARIADB_PASSWORD=insecure
+MARIADB_ROOT_PASSWORD=insecure
+
+## There's probably very little reason to change any values below here.
+# authentication mode (public, password)
+PHOTOPRISM_AUTH_MODE=password
+# disables HTTPS/TLS even if the site URL starts with https://
+PHOTOPRISM_DISABLE_TLS=true
```
EnigmaCurry, line 62: this should not be exposed, because it should always be set since its behind traefik.

photoprism/docker-compose.yaml
mcmikemn: I changed to `no-new-priveleges:true` and it seems to work fine.

photoprism/.env-dist
mcmikemn: it seems this might be cosmetic, but I corrected it anyway

photoprism/.env-dist
mcmikemn: fixed

photoprism/Makefile
mcmikemn: fixed

photoprism/.env-dist
mcmikemn: fixed

photoprism/.env-dist
mcmikemn: fixed

photoprism/.env-dist
mcmikemn: fixed

mcmikemn:
I request that you check my fixes to make sure I fixed them in a decent way and didn't introduce other issues.

photoprism/Makefile
```diff
@@ -11,13 +11,14 @@ config-hook:
...
 	@${BIN}/reconfigure ${ENV_FILE} PHOTOPRISM_INSTANCE=$${instance:-default}
 	@${BIN}/reconfigure_htpasswd ${ENV_FILE} PHOTOPRISM_HTTP_AUTH default=no
-	
+	@${BIN}/reconfigure_password ${ENV_FILE} PHOTOPRISM_DATABASE_PASSWORD 32
+
 	@echo ""
 	@${BIN}/reconfigure_ask ${ENV_FILE} PHOTOPRISM_ADMIN_USER "Enter the admin user name" ${PHOTOPRISM_ADMIN_USER}
 	@echo ""
-	@${BIN}/reconfigure_ask ${ENV_FILE} PHOTOPRISM_ADMIN_PASSWORD "Enter the admin password" ${PHOTOPRISM_ADMIN_PASSWORD}
+#	@${BIN}/reconfigure_ask ${ENV_FILE} PHOTOPRISM_ADMIN_PASSWORD "Enter the admin password" ${PHOTOPRISM_ADMIN_PASSWORD}
```
EnigmaCurry, line 19: I'm thinking that if the admin password is an initial password only, and that the user needs to change it after its loaded, then this can't really be changed later from the .env, so it should just be a static password (ie 'password'), documented, and tell the user to change it in the interface instead. Otherwise its confusing that the password is here in the .env but it can't really be changed here. (If I understand it correctly)

photoprism/docker-compose.yaml
```diff
@@ -33,18 +33,18 @@ services:
...
     #ports:
     #  - "2342:2342" # HTTP port (host:container)
     environment:
       - PHOTOPRISM_ADMIN_USER
       - PHOTOPRISM_ADMIN_PASSWORD
       - PHOTOPRISM_AUTH_MODE
-      - PHOTOPRISM_SITE_URL
-      - PHOTOPRISM_DISABLE_TLS
+      - PHOTOPRISM_SITE_URL=https://${PHOTOPRISM_TRAEFIK_HOST}
+      - PHOTOPRISM_DISABLE_TLS=false
```
EnigmaCurry, line 47: I think DISABLE_TLS should be true. Because traefik will be adding the TLS on top.

EnigmaCurry (approved):

photoprism/Makefile
mcmikemn: Yeah, you got it right. I changed it how you suggested.

photoprism/docker-compose.yaml
mcmikemn: Fixed.

EnigmaCurry (approved):

PR #86, opened 2023-07-27
added hmac_key

mcmikemn:
I tried to install Invidious but the container remained unhealthy. Logs showed `Config: 'hmac_key' is required/can't be empty`, and I found [this](https://github.com/iv-org/invidious/blob/d956b1826e15da6cfcd9a1531b0f1e6ef577dd10/config/config.example.yml#L438-L468) and [this](https://docs.invidious.io/installation/#docker-compose-method-production) (search for "hmac"). I figured it was a new development.


EnigmaCurry (approved):

EnigmaCurry:
Thanks, yes it appears new (or newly required) https://github.com/iv-org/invidious/issues/3854

Issue #88, opened 2023-07-29
instance oddities in d.rymcg.tech


mcmikemn:
After I create an instance and `<ctrl>-d` to return to the main shell, if I type `make switch`, d.rymcg.tech puts me in a subshell for the default instance - it doesn't ask me what instance I want to switch to. The same thing happens if I type `make switch <instance_name>`.

If I type `make instance`, d.rymcg.tech lists all instances of the projet and prompts me to enter one, then puts me in a subshell for that instance - this is great!

EnigmaCurry:
It works this way: `make switch instance=foo`

EnigmaCurry:
a bit more backstory: make doesn't allow positional arguments, it only allows named arguments. So any of these work and do the same thing:

`export INSTANCE=foo; make switch`
`INSTANCE=foo make switch`
`make INSTANCE=foo switch`
`make switch INSTANCE=foo`

`INSTANCE` and `instance` may be used interchangeably.

mcmikemn:
Changed README to clarify in [PR #90](https://github.com/EnigmaCurry/d.rymcg.tech/pull/90)

PR #90, opened 2023-07-30
clarify `make switch` in readme

mcmikemn:
This sentence in README confused me re: `make switch` 
> If you want to enter the sub-shell without automatically running `make config`, you can run `make switch` rather than `make instance`.

So I updated it to
> If you want to enter the sub-shell without automatically running `make config`, you can run `make switch instance=foo` rather than `make instance`.

This fixes [issue #88](https://github.com/EnigmaCurry/d.rymcg.tech/issues/88).

### DIFF
--- a/README.md
+++ b/README.md
@@ -852,7 +852,8 @@ To exit the sub-shell, press `Ctrl-D` or type `exit` and you will
 return to the original parent shell and working directory.
 
 If you want to enter the sub-shell without automatically running `make
-config`, you can run `make switch` rather than `make instance`.
+config`, you can run `make switch instance=foo` rather than
+`make instance`.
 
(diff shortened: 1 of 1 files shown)


EnigmaCurry:
Here on githb if you say: `This fixes #88` - instead of "addresses" it will link it up so that when this is merged it automatically closes the #88.

This fixes #88

mcmikemn:
> Here on githb if you say: `This fixes #88` - instead of "addresses" it will link it up so that when this is merged it automatically closes the #88.
> 
> (I think it only works in the first description not follow up comments)

Good to know! I edited my first comment, in case that re-establishes the connection between the merge and this issue.

PR #92, opened 2023-07-31
Lemmy

mcmikemn:
After @EnigmaCurry replaced the original nginx service with a custom nginx build, everything is working.


lemmy/.env-dist
```diff
@@ -0,0 +1,32 @@
...
+LEMMY_TRAEFIK_HOST=lemmy.example.com
+LEMMY_VERSION=0.18.3
+LEMMY_UI_VERSION=0.18.3
+ASONIX_PICTRS_VERSION=0.4.1
+
+# The name of this instance. If there is only one instance, use 'default'.
+LEMMY_INSTANCE=
+
+# You can degine the port that Traefik uses for the lemmy service.
+LEMMY_NIGNX_PORT=8536
```
EnigmaCurry, line 11: misspelled. Also, I don't think this needs to be exposed.

EnigmaCurry: It's static in the nginx.conf so it cant be configured otherwise.

mcmikemn:
Oh, and I have to delete orig-compose.yml

lemmy/.env-dist
mcmikemn: fixed in [494d31e](https://github.com/EnigmaCurry/d.rymcg.tech/pull/92/commits/494d31e20d60eddb535e1834d1b8b4e8a65d0d3e)

lemmy/docker-compose.yaml
```diff
@@ -0,0 +1,132 @@
...
+
+  pictrs:
+    image: asonix/pictrs:${ASONIX_PICTRS_VERSION}
+    # this needs to match the pictrs url in lemmy.hjson
+    hostname: pictrs
+    # we can set options to pictrs like this, here we set max. image size and forced format for conversion
+    # entrypoint: /sbin/tini -- /usr/local/bin/pict-rs -p /mnt -m 4 --image-format webp
+    environment:
+# comment out the telemetry line?
+      - PICTRS_OPENTELEMETRY_URL=http://otel:4137
```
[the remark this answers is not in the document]
mcmikemn: Lemmy seems to work fine without it. Removed it in [494d31e](https://github.com/EnigmaCurry/d.rymcg.tech/pull/92/commits/494d31e20d60eddb535e1834d1b8b4e8a65d0d3e)

lemmy/docker-compose.yaml
```diff
@@ -21,7 +21,7 @@ services:
       # actual and only port facing any connection from outside
       # Note, change the left number if port 1236 is already in use on your system
       # You could use port 80 if you won't use a reverse proxy
-      - "${LEMMY_NIGNX_PORT}:8536"
+      - "8536:8536"
```
EnigmaCurry, line 24: there should be no exposed port, traefik should do the work.

mcmikemn: I just pushed that fix. What are you viewing?

EnigmaCurry: remove the whole ports section.

EnigmaCurry: any port listed in the port section means to bypass traefik completely, we don't want that.

mcmikemn: Oohhh! I misunderstood what you meant by "exposed." Sorry. Fixed in [ebe85e9](https://github.com/EnigmaCurry/d.rymcg.tech/pull/92/commits/fe8065f2f2b206a2f419e66e6a33ef527ebe85e9)

EnigmaCurry:
did you test `make instance`?

lemmy/.env-dist
```diff
@@ -0,0 +1,29 @@
+# The domain name for the whoami service:
+LEMMY_TRAEFIK_HOST=lemmy.example.com
+LEMMY_VERSION=0.18.3
+LEMMY_UI_VERSION=0.18.3
+ASONIX_PICTRS_VERSION=0.4.1
```
EnigmaCurry, line 5: This var should start with LEMMY_

lemmy/README.md
```diff
@@ -0,0 +1,40 @@
...
+`passwords.json`).
+
+## Destroy
+
+```
+make destroy
+```
+
+This completely removes the container (and would also delete all its
+volumes; but `whoami` hasn't got any data to store.)
```
EnigmaCurry, line 40: remove this text about whoami

lemmy/.env-dist
mcmikemn: I thought about that. I chose not to because it's pulling a non-Lemmy service even though it's in d.rymcg.tech/lemmy. But now that I think about it more, we do that all the time - so I'll change it.

EnigmaCurry: Yea the prefix is indicating which d.ry project the var is for, lemmy.

EnigmaCurry: I've never tested it, but d.ry is designed so that all the vars could go in one giant .env file, and there would be no naming conflicts because they are prefixed

lemmy/nginx/nginx.conf
```diff
@@ -0,0 +1,95 @@
...
+        client_max_body_size 20M;
+
+        add_header X-Frame-Options SAMEORIGIN;
+        add_header X-Content-Type-Options nosniff;
+        add_header X-XSS-Protection "1; mode=block";
+
+        # Send actual client IP upstream
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header Host $host;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
```
EnigmaCurry, line 72: This part may be broken (or might not) because traefik is proxy in fron if nginx. Traefik already does this and sets these headers, but nginx is probably not allowing/trusting that. So if lemmy does something with the X-Real-IP header, it might be broken.

mcmikemn: I don't know how to test this.

lemmy/.env-dist
mcmikemn: fixed in [21d70f3](https://github.com/EnigmaCurry/d.rymcg.tech/pull/92/commits/21d70f3d793f066ab97edbf562415b039fefc6dc)

lemmy/README.md
mcmikemn: fixed in [e481bc5](https://github.com/EnigmaCurry/d.rymcg.tech/pull/92/commits/e481bc5554f44a104e3da905eb6b862e41540bf7)

mcmikemn:
> did you test `make instance`?

Yes.

lemmy/nginx/nginx.conf
EnigmaCurry: it might not be a problem, but it does look like they might be using it for something because they went through the trouble of adding it. If its broken, then the X-Real-IP won't be showing the ip address of the user, but instead it will show the IP address of traefik (not super useful).

I can think of two ways to test it: 
 * Log the request and header from lemmy somehow.
 * Replace lemmy frontend with a whoami container, whoami always prints the headers, including X-Real-IP and X-Fowarded-For, which if it works, should show your browser's ip address, not traefik ip address.

mcmikemn: Are such headers viewable in the browser DevTools?

EnigmaCurry: they are not, because these are the headers that come from traefik to nginx, your browser is no longer a party to that conversation.

EnigmaCurry: I assume lemmy is using this feature to record the real ip address of posts. so can you see it on one of the posts metadata?

EnigmaCurry: I tested it by replacing the lemmy-ui container with this:

```
  lemmy-ui:
    image: traefik/whoami
    command: --port 1234
    depends_on:
      - lemmy
    security_opt:
      - no-new-privileges:true
    restart: unless-stopped
```

And then nginx forwards to the whoami, when I load it returns this:

```
Hostname: d8928ea3e027
IP: 127.0.0.1
IP: 172.25.7.3
RemoteAddr: 172.25.7.8:45964
GET / HTTP/1.1
Host: lemmy.maldek.nogn.rymcg.tech
User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/115.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US,en;q=0.5
Connection: close
Dnt: 1
Sec-Fetch-Dest: document
Sec-Fetch-Mode: navigate
Sec-Fetch-Site: none
Sec-Fetch-User: ?1
Upgrade-Insecure-Requests: 1
X-Forwarded-For: 10.13.13.86, 172.25.7.1
X-Forwarded-Host: lemmy.maldek.nogn.rymcg.tech:8443
X-Forwarded-Port: 8443
X-Forwarded-Proto: https
X-Forwarded-Server: docker-vm
X-Real-Ip: 172.25.7.1
```

which you see at the bottom X-Real-Ip, thats the IP address of my docker host where traefik is running (on the host network) so its not returning correctly the users IP address.

EnigmaCurry: Actually, X-Forward-For *is* working, although X-Real-Ip is not.

EnigmaCurry: I guess I wont worry about for now

PR #94, opened 2023-08-02
`make instance` source RC file

### LINKED ISSUE #89: `make instance` and `make switch` should optionally source ~/.bashrc
### Discussed in https://github.com/EnigmaCurry/d.rymcg.tech/discussions/87

<div type='discussions-op-text'>

<sup>Originally posted by **mcmikemn** July 29, 2023</sup>
When you `make instance`, d.rymcg.tech creates a sub-shell for the new instance. Can that subshell retain or copy the environment from the main shell? For example, I have aliases set up in .bashrc in my main shell, but they're not available in the sub-shell.</div>

Because we cannot control what a user has in their ~/.bashrc it should not be the default behaviour to source it in the `make instance` subshell, however it might be nice to extend the environment by sourcing a select file by configuration.

A root environment variable like `MAKE_INSTANCE_SOURCE_RCFILE=` would be blank (or `/dev/null`) by default, but if you set this variable to a file path, it should source that file it before the other variables in the [make instance/switch](https://github.com/EnigmaCurry/d.rymcg.tech/blob/master/_scripts/Makefile.instance) script, thus extending the user's environment.

mcmikemn:
trying to source MAKE_INSTANCE_RC_FILE in expression passed to `bash --rcfile <(...) -i`

Fixes #89


EnigmaCurry:
Thanks mike, I fixed the file existence test, and I set the default to blank, which follows the way we handle other defaults.

EnigmaCurry:
@mcmikemn I think your way was close, maybe you just needed an && and the end of it before the next echo, but its cleaner this way if its all one string instead of two.

mcmikemn:
Sheesh. Way simpler than I was making it. Thanks!

EnigmaCurry:
If it works for you, please (squash) merge this

PR #96, opened 2023-08-03
Switch menu

mcmikemn:
I like the instance-selection menu that `make install` had, so I added it to `make switch`, too.


EnigmaCurry:
Thanks Mike, I like your addition, but at the same time I want to preserve some of the existing functionality, specifically if the user types `make switch instance=foo` it should not prompt for which instance to switch to.

EnigmaCurry:
I see you added the new script `switch` - I think we ought to be able to have only one script either `instance` or `switch`, and one or the other should be able to be  implemented in terms of the other.

EnigmaCurry:
I want `make instance` to always rerun the config, because the Makefile may have changed, and therefore the config needs to be updated. I guess thats not going to happen too often for the user who only cares for the abstraction and will never change the internals; but for any developer this is important so that they don't forget to run it. Thats the reason I made `make switch` for cases where I know what I'm doing and I know my config is up to date.

changes needed to `make instance`

 * `make instance instance=foo` should not prompt for the instance, but should run the config for foo.

`make switch` should be identical to `make instance`:

 * except `make switch` doesn't run the config, even if its missing.
 * except `make switch` should prompt for the instance if its not specified.
 * `make switch instance=foo` should not prompt for the instance, just like `make instance instance=foo` should.

mcmikemn:
> changes needed to `make instance`
> 
>     * `make instance instance=foo` should not prompt for the instance, but should run the config for foo.
> 
> 
> `make switch` should be identical to `make instance`:
> 
>     * except `make switch` doesn't run the config, even if its missing.
> 
>     * `make switch` should prompt for the instance if its not specified, just like `make instance` should.
> 
>     * `make switch instance=foo` should not prompt for the instance, just like `make instance instance=foo` should.

I like this idea. I'll work on it.

EnigmaCurry (approved):

EnigmaCurry:
Looks great @mcmikemn thanks!

PR #99, opened 2023-08-16
Docker socket

### LINKED ISSUE #72: Support Docker in Docker using /var/run/docker.sock
Currently only remote SSH contexts are supported by d.rymcg.tech. It is considered an anti-pattern to install it directly to the same machine that runs the Docker daemon (always clone it to your workstation instead). However, in the case of a containerized "workstation", using Docker in Docker, the usual way is to bind mount /var/run/docker.sock, so it would be nice to be able to support that as well.

EnigmaCurry:
This branch enables using Docker via the unix socket /var/run/docker.sock instead of the default SSH context.

Fixes #72 

Things that are still broken:
 * traefik `make open` needs SSH still
 * traefik cannot use host networking on Docker Desktop.

### DIFF
--- a/traefik/Makefile
+++ b/traefik/Makefile
@@ -11,7 +11,7 @@ config-hook: traefik-user config-ask compose-profiles
 
 .PHONY: traefik-user
 traefik-user:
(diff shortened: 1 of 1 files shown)


EnigmaCurry:
Even with the drawbacks it still fixes #72 as written, and looks like it didnt break the degault ssh context config either so I'll merge this.

PR #104, opened 2023-08-18
Podfetch

mcmikemn:
Added Podfetch. It "works", but I can't even figure out how their built-in auth functions! [Docs](https://github.com/SamTV12345/PodFetch/blob/main/docs/AUTH.md) say to set env var `BASIC_AUTH` to false to turn off their built-in auth, but even with `PODFETCH_BASIC_AUTH=false` in .env*, as long as the env vars `USERNAME` and `PASSWORD` are set, Podfetch's internal auth turns itself on.

And with their internal auth on, I can't add a podcast! When I try to add one from an iTunes search or import an OPML file, I get "not authorized" errors (in the UI or in the log, respectively).

If we decide to turn off their internal auth (by removing the env vars from `compose.yaml`) and use Traefik's HTTP Basic Auth instead, I think the only way to add/manage podcasts is through their CLI. This would be fine, and we could probably add functionality to `Makefile` to incorporate these commands.


EnigmaCurry:
you should use the docker-compose.instance.yaml template to conditionally add the env vars depending on whether PODFETCH_BASIC_AUTH is set to `true` or not, that way you only add them when its right to.

So the Makefile should guide you to picking http basic auth OR the podfetch basic auth, and turn the other feature off automatically (they should be mutually exclusive).

The PODFETCH_PASSWORD should be randomized in the Makefile, or you can get fancy and ask for it and only randomize it if its left blank.

EnigmaCurry:
in fact there should be three auth choices:

 * traefik http basic auth
 * podfetch "basic auth"

EnigmaCurry:
https://github.com/SamTV12345/PodFetch/blob/d9c52def75fb3566f1cf12127d929a9befb09fe4/src/auth_middleware.rs#L77

```
        return if var(BASIC_AUTH).is_ok() {
```

`is_ok()` here is not checking if the value is `true`, but rather its just checking for its existance, therefore even if the value is `false` the feature is enabled.

mcmikemn:
> you should use the docker-compose.instance.yaml template to conditionally add the env vars depending on whether PODFETCH_BASIC_AUTH is set to `true` or not, that way you only add them when its right to.
> 
> So the Makefile should guide you to picking http basic auth OR the podfetch basic auth, and turn the other feature off automatically (they should be mutually exclusive).
> 
> The PODFETCH_PASSWORD should be randomized in the Makefile, or you can get fancy and ask for it and only randomize it if its left blank.

I made these changes. Unrelated, I still can't add a podcast. :) When I try (even logged in as a podfetch Admin user), the podfetch log says `IO error: Permission denied (os error 13)`. Maybe this is an issue with the container not being able to write to the volume?

EnigmaCurry:
I updated this to use the new auth config

I had to disable the gpodder api feature, because it apparently has a hard requirement of handling the auth itself. So can it be made to handle the X-Forwarded-User from traefik-forward-auth?

PR #111, opened 2023-09-17
make Homepage autoconfigure by default, or use config repo based on .env var

mcmikemn:
Default: `HOMEPAGE_AUTO_CONFIG=true` and `make config` will configure Homepage for all apps currently running in current Docker context.

`HOMEPAGE_AUTO_CONFIG=false` and `make config` works as it already does, pulling Homepage config from repo.

I did it all in bash instead of using ytt. Autoconfig is pretty limited unless we want to go deeper with logic to determine which of an apps containers is the app the user wants to show container stats for in Homepage, allow the user to change the order or the apps (or rename them, or add a description, or change the appearance of the dashboard, etc.). I don't think that's necessary for the intent of autoconfig (as I understand it): make a nice dashboard for all d.ry apps, automatically, easily. If a user wants to configure it more, they can fork the config repo and go that route.

Your thoughts?


EnigmaCurry (approved):
Thanks @mcmikemn I finally got around to test testing this, looks good!

PR #113, opened 2023-09-29
added `git branch` to completions

mcmikemn:
I added a completion for `branch`:
`d.rymcg.tech branch` lists all local git branches
`d.rymcg.tech branch X` switches to that branch

### DIFF
--- a/_scripts/d.rymcg.tech
+++ b/_scripts/d.rymcg.tech
@@ -183,6 +183,14 @@ __context() {
     fi
 }
 
+__branch() {
+    if [[ $# == 0 ]]; then
+        exe git branch
+    else
+        exe git checkout "$@"
+    fi        
+}
+
 __run_script() {
     CMD="${1}"; shift
     ${ROOT_DIR}/_scripts/${CMD} "$@"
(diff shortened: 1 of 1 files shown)


EnigmaCurry:
This looks convenient, do you foresee adding additional git commands besides branch?

What if instead we added a `git` subcommand, that just deferred to using git in the d.ry ROOT_DIR?

So it would be just like git except you wouldn't have to change directories: 

```
d.rymcg.tech git branch
d.rymcg.tech git checkout whatever
d.rymcg.tech git commit ....
```

You can get it to defer to the git autocompletion too so it completes any git command.
I mostly always use magit (as opposed to running git in the shell) so I can't envision using this myself.

mcmikemn:
I like the idea of adding a `git` subcommand, but my first thought is: it wouldn't add convenience, it would add _**in**_convenience by having to type `d.rymcg.tech` first. But `d.rymcg.tech` can be autocompleted, so that makes the added inconvenience tiny. But then if all git commands could be autocompleted, that pushes the qhole equation back on the side of convenience. Not much, but there's also the "cool" factor. :)

mcmikemn:
I doubt I'd use the git subcommand completion much either, actually. My git use is pretty limited - short and easy commands. It's cool, but neither of us really plan to use it.

EnigmaCurry:
`d.rymcg.tech git branch` is a lot easier for me to type than `cd ~/git/vendor/enigmacurry/d.rymcg.tech; git branch`.. But `Ctrl-c g` to open magit is quicker still in emacs.

mcmikemn:
I don't foresee needing to check or change the git branch unless I'm also going to be in `~/git/vendor/enigmacurry/d.rymcg.tech`. It's nice that the `d.rymcg.tech` completion can happen anywhere, but I'm only ever going to use it from within d.ry directories anyway.

But theoretically, someday, maybe, there may be others who use d.ry! :)

At the moment, I'm not interested in trying to figure out how to add `git` as a subcommand and defer to the git autocompletion, but I'm not opposed to the idea.

EnigmaCurry:
I prefer not overriding the branch changing syntax (`branch X` vs `checkout X`) and also it feels a bit specific to only do `branch`, so I am in favor of adding the git subcommand over this. Thanks for the suggestion!

I opened #114

Issue #118, opened 2023-10-08
add authorization middlewares to go along with TFA


mcmikemn:
Currently, you can install Gitea and use it for authentication, and if you install traefik-forward-auth and customize an app to use it, Gitea will require a login to access your app. But it will not affect what a successfully logged-in person can do in the app. If the app has built-in authentication/authorization and can check the user header that traefik-forward-auth sends, then the app can limit what the logged-in person can do in the app. But if the app can't check the user header, or if the app doesn't have built-in authentication/authorization at all, then any person with an account on your Gitea server can log into the app and have full access. And any user with an account on your Gitea server can log into **_all_** of your apps that have traefik-forward-auth configured, not just a select one or few.

One way to improve this is to add/create a middleware that receives the user header from traefik-forward-auth and compares it to a list of usernames that are allowed to log into the app; if the username from traefik-forward-auth is in the approved list, they can log into the app - if it isn't, they receive a 401. The list of approved usernames would be configurable in the app's .env file. The result would be that a user on your Gitea server could log into only those apps that are configured to allow them to log in; however, if those apps can't check the user header or don't have built-in authentication/authorization at all, then the user still has full access to the app once they're logged in. This might not be too big of a deal, because if an app doesn't have it's own internal authentication/authorization then there are no "permissions" to speak of - if you can use the app, you can use the app. But if an app does have it's own internal authentication/authorization but can't check the user header from traefik-forward-auth, then will all users who can log in have the same access?

EnigmaCurry:
I think this can be done with a Traefik plugin like this: https://github.com/poloyacero/headauth

A plugin is a special kind of middleware that is builtin to Traefik, so like the blockpath and geoip plugins, this would be installed into traefik itself, and not require another middleware container to be started.

EnigmaCurry:
I forked it here https://github.com/EnigmaCurry/traefik-header-authorization. I will test installing it in the traefik image.

PR #120, opened 2023-10-10
Header Authorization

### LINKED ISSUE #125: Fix qbittorrent
when I install it I get `Unauthorized` because its setting a password for the interface, but there is no associated env var that sets it.

Fixed in #120

### LINKED ISSUE #128: `make clean` and especially `make clean-all` should check for existing instances beforehand
`make clean` currently only handles the cleanup of the .env files and passwords.json files.

If you run `make clean` before removing the running instance, you end up with an instance that is harder to cleanup, as `make destroy` requires the .env file in order to function. 

So `make clean` should error and alert the user (run `make ps` to show the user the one existing instance) if it detects its existing instance container/volume.

And `make clean-all` should error and alert the user (run `make status` to show the user all of them) if it detects *any* existing instance container/volume.

EnigmaCurry:
This PR will track implementing #118 which will add header based authorization to apps that don't normally support being hosted behind an oauth proxy, or to provide an additional layer of authorization. 

DONE: 
 * Installed traefik plugin to act as middleware after traefik-forward-auth to filter approved usernames - https://github.com/EnigmaCurry/traefik-header-authorization
 * added whoami example

TODO:
* configure all the .env-dist, Makefile, and docker-compose.instance.yaml for all apps that benefit from t-f-a.

Fixes #125 
Fixes #128


EnigmaCurry:
I added a hardcoded example to whoami and tested that it works. This needs to be made configurable from the env file and the reconfigure_oauth script needs to be modified to ask for the list of the allowed users (gitea email addresses). Do you want to test it and take over since you already made that script @mcmikemn ?

mcmikemn:
> Do you want to test it and take over since you already made that script @mcmikemn ?

Sure! Working on it...

mcmikemn (changes requested):
In the .env-dist files, instead of saying "With Oauth2 authentication, you can configure which email addresses...", make it more clear that the basic authorization can be configured but is not part of Oauth2 authentication. Maybe "In addition to Oauth2 authentication, you can configure basic authorization by entering which email addresses..."

mcmikemn (dismissed):

mcmikemn:
@EnigmaCurry, this looks good to me and I tested it with whoami and homepage.

EnigmaCurry:
I think it would be nicer if it it offered a trinary response "Do you want to enable authentication?" 1) No 2) HTTP Basic 3) Oauth2 

I know we have no examples of that in the makefiles up to this point. I can't imagine why anyone would mix basic and oauth at the same time.

EnigmaCurry:
well it could work like this:

Do you want to enable traefik authentication (in addition to any app auth)? N/y

If yes, detect if t-f-a is installed, and ask "Do you want to use Oauth2" and if no or not installed, ask "Do you want to use HTTP Basic Auth". 

No trinary response needed if its broken up into multiple y/n questions.

EnigmaCurry:
I moved the middleware config to the Traefik config, and added a script `make groups` to manage the group config.

I modified the whoami example with a hardcoded header-authorization-group middleware for the `users` group.

We still need a script that the app Makefile can run to select which kind of auth the user wants, and if they do want oauth2 it needs to set the correct middleware for the group. (To keep things simple, I think the question should be worded about what group the user wants to apply (ie. `users`)  not which middleware (ie. not `header-authorization-group-users`)

mcmikemn:
@EnigmaCurry, I'm adding the "remove user from group" function to `_scripts/reconfigure-header-authorization-groups`, but I can't get 1 line to work: https://github.com/EnigmaCurry/d.rymcg.tech/blob/67cffb179a54a78a71c4832994c458db85392bce/_scripts/reconfigure_header_authorization_groups#L125
It errors with "Cannot iterate over null."

I also added display info to the other functions to make this script easier to use. And I added "enter blank to skip" to the function calls, because otherwise the only way out of a command was to <ctrl>-C the script.

EnigmaCurry:
I couldn't figure out how to remove all the selected users in one line so I had to use a loop

https://github.com/EnigmaCurry/d.rymcg.tech/blob/1960e9504d6d835d666f6a93735a0891bb6fce89/_scripts/reconfigure_header_authorization_groups#L132-L135

mcmikemn:
I combined all auth config questions, made authroization group configurable in docker-compose.instance.yaml, added reconfigure_menu for general use (I didn't incporate it into `reconfigure_header_authorization_groups`, but it would be a good use for it), updated readmes.

I only modified and tested whoami - if this works well and is the way we want to go, I still need to update each app that will have oauth2 as an option (audiobookshelf, drawio, grocy, homepage, invidious, lemmy (by removing the option to add external authentication), photoprism, piwigo, wordpress).

_scripts/reconfigure_auth
```diff
@@ -0,0 +1,52 @@
...
+## Subsume all the authentication and authorization configurations in 1 script
+
+BIN=$(dirname ${BASH_SOURCE})
+source ${BIN}/funcs.sh
+set -eo pipefail
+
+ENV_FILE=$1
+VAR_PREFIX=$2
+
+## note: should we add comments to reconfigure_htpasswd and reconfigure_oauth2 indicating they were designed to be called by this script, not independently?
```
EnigmaCurry, line 13: We can set some crazy env var in this script like __D_RY_CONFIG_ENTRY=reconfigure_auth and then if that variable is not set correctly in reconfigure_htpasswd or reconfigure_oauth2, then it aborts.

EnigmaCurry: yea I think thats the idea.

EnigmaCurry:
Thanks mike! 

 * Please use four spaces instead of tabs in scripts.
 * Do not `source` scripts to return values. You need to print and parse instead. 
 * I would recommend using stderr for human text and stdout for the response text (which for multiple values, needs to be a JSON response, and parsed from the caller)

EnigmaCurry:
```
OPTIONS=$1
PROMPT=$2
DEFAULT=$3
MIN=${4:-1}
MAX={5:-$MIN}
check_var OPTIONS PROMPT DEFAULT
check_num MIN MAX
```
you can put something like this at the top to check the correct args are set before you start processing them.

EnigmaCurry:
I have merged the script-wizard branch into the header-authorization branch.

mcmikemn:
I updated `reconfigure_auth` to incorporate script-wizard. It's only functional in `whoami` at the moment. Once it's good, I'll update the other apps that benefit from t-f-a.

whoami/.env-dist
```diff
@@ -27,7 +27,7 @@ WHOAMI_HTTP_AUTH=
 WHOAMI_OAUTH2=no
 # In addition to Oauth2 authentication, you can configure basic authorization
 # by entering which authorization group can log into your app. You create
-# groups of email addresses in the `traefik` folder by running `makr groups`. 
+# groups of email addresses in the `traefik` folder by running `make groups`. 
 WHOAMI_AUTHORIZED_GROUP=
```
EnigmaCurry, line 31: I think this should be WHOAMI_OAUTH2_AUTHORIZED_GROUP

mcmikemn: Fixed.

EnigmaCurry:
Added confirmation before installing script-wizard.

_scripts/reconfigure_oauth2
```diff
@@ -12,9 +12,13 @@ set -e
 ENV_FILE=${1}
 shift
 
+## env var for oauth2
 VAR=${1}
 shift
 
+## env var for authorized emails
+VAR2=${2}
```
EnigmaCurry, line 20: Something's wrong with the args here. You use `shift` (which pops the first arg off and the second arg becomes $1 for each time you call `shift`), so it shouldn't be $2 here. When I changed it to $1 and I inspected the value it was `VAR=default=yes` which also seems wrong.

EnigmaCurry: These should not be called `VAR` and `VAR2` please pick descriptive names.

_scripts/reconfigure_oauth2
```diff
@@ -35,9 +39,28 @@ if [[ ${1} =~ ^default= ]]; then
...
-    ${BIN}/confirm ${DEFAULT_ENABLED_OAUTH2} "Do you want to enable OpenID/OAuth2" "?" || disable_oauth2
+    echo 
+    echo "Using OpenID/OAuth2 will require a login to access your app. You can configure basic authorization by assigning an authorization group (authorization groupscontain email addresses that are allowed to log into your app - you configure authorization groups by running \`make groups\` in the \`traefik\` directory)."
+    echo 
+    echo "SECURITY NOTE: Using OpenID/OAuth2 is on top of any authentication/authorization service your app provides. OpenID/Oauth2 will require a login to access your app and permit only specific logins, but it will not affect what a successfully logged-in person can do in your app. If your app has a built-in authorization mechanism that can check for the user header that traefik-forward-auth sends, then your app can limit what the logged-in person can do in the app. But if your app can't check the user header, or if your app doesn't have built-in authorization at all, then any person with an account on your Gitea server that you permit to log into your app will have full access."
```
EnigmaCurry, line 45: I think the security note is out of date assuming that we enforcing the groups, its not a concern anymore.

_scripts/reconfigure_oauth2
mcmikemn: $2 after the shift was correct because $1 is used elsewhere, but I'll rename the vars

_scripts/reconfigure_oauth2
mcmikemn: You want me to remove the entire paragraph? from `reconfigure_oauth2` (i.e., the prompt when you're starting to configure oauth2), from the app readmes, and from the t-f-a readme?

_scripts/reconfigure_oauth2
mcmikemn: fixed in upcoming commit

_scripts/reconfigure_auth
mcmikemn: added.

_scripts/reconfigure_oauth2
EnigmaCurry: how about replacing it with:

OAuth2 uses traefik-forward-auth to delegate authentication to an external authority (eg. a self-deployed gitea instance). Accessing this app will require all users to login through that external service first. Once authenticated, you may be authorized access only if your login id matches the member list of the predefined security group configured for the app (`WHOAMI_OAUTH2_AUTHORIZED_GROUP`) . Security groups are defined in the Traefik config (`TRAEFIK_HEADER_AUTHORIZATION_GROUPS`) and can be modified using `make groups` (in the traefik directory).

EnigmaCurry:
reconfigure_oauth2 now checks if traefik-forward-auth is installed.

_scripts/reconfigure_oauth2
EnigmaCurry: > You want me to remove the entire paragraph? from `reconfigure_oauth2` (i.e., the prompt when you're starting to configure oauth2), from the app readmes, and from the t-f-a readme?

I think its probably good to have it in each app readme, but specifically we dont need to worry about telling the user that "any user can do whatever they want" because now they have to be a part of a group.

_scripts/reconfigure_oauth2
```diff
@@ -9,22 +9,26 @@ BIN=$(dirname ${BASH_SOURCE})
 source ${BIN}/funcs.sh
 set -e
 
+if [[ __D_RY_CONFIG_ENTRY != reconfigure_auth ]]; then
```
EnigmaCurry, line 12: ${__D_RY_CONFIG_ENTRY}

_scripts/reconfigure_htpasswd
```diff
@@ -9,6 +9,12 @@ BIN=$(dirname ${BASH_SOURCE})
 source ${BIN}/funcs.sh
 set -e
 
+if [[ __D_RY_CONFIG_ENTRY != reconfigure_auth ]]; then
```
EnigmaCurry, line 12: ${__D_RY_CONFIG_ENTRY}

mcmikemn: fixed

_scripts/reconfigure_oauth2
mcmikemn: updated.

_scripts/reconfigure_oauth2
mcmikemn: fixed.

piwigo/README.md
```diff
@@ -29,8 +19,30 @@ Alternatively to running `make config`, you can manually copy
  * `MARIADB_PASSWORD` the mysql user password
  * `TIMEZONE` the timezone, in the format of [TZ database name](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones)
  * `PIWIGO_OAUTH2` set to `yes` to configure Oauth2 authentication (in addition to Piwigo's own internal authentication)
+ * `PIWIGO_AUTHORIZED_IDENTITIES` enter a comma-separated list of email addresses authorized to log into this app (must match Gitea accounts)
```
mcmikemn, line 22: this needs to reflect groups, not emails

traefik-forward-auth/README.md
```diff
@@ -85,31 +100,40 @@ See the [whoami](../whoami/) app for examples.
...
       #@ if enable_oauth2 == "yes":
       #@ enabled_middlewares.append("traefik-forward-auth@docker")
+      #@ enabled_middlewares.append("header-authorization-group-{}@file".format(authorized_group))
       #@ end
 ```
 
 ## Logging out
 
-User logout is a multi-phase endevour:
+User logout is a multi-phase endeavour:
```
mcmikemn, line 136: spelling: "endeavor"

mcmikemn (changes requested):
all app readmes refer to WHOAMI* env var - change that to refer to each app's env var.

piwigo/README.md
mcmikemn: fixed.

mcmikemn (approved):

EnigmaCurry:
> Do you want to rename `confirm.sh` to `confirm_script_wizard.sh`, and `confirm` to `confirm.sh`?

My intention was to replace the old bash `confirm` and `ask` with the new script-wizard equivalents. Which, the only nicer thing about them I can see is that it normalizes the printed response as `Yes` or `No` in blue text. (I already backported the green ? prompt.)

So i easily replaced confirm, but I still need the old one as confirmation before installing script-wizard in the first place.

I tried replacing ask but i couldn't replicate the "response on a new line" feature we have now, so i did not replace ask for now.

As far as the naming, I like confirm.sh because it makes it clear it is bash. It doesn't really matter which one you use they function nearly identically, but I would prefer to use the new one `confirm`. I hope its not confusing.

mcmikemn:
> > Do you want to rename `confirm.sh` to `confirm_script_wizard.sh`, and `confirm` to `confirm.sh`?
> 
> My intention was to replace the old bash `confirm` and `ask` with the new script-wizard equivalents. Which, the only nicer thing about them I can see is that it normalizes the printed response as `Yes` or `No` in blue text. (I already backported the green ? prompt.)
> 
> So i easily replaced confirm, but I still need the old one as confirmation before installing script-wizard in the first place.
> 
> I tried replacing ask but i couldn't replicate the "response on a new line" feature we have now, so i did not replace ask for now.
> 
> As far as the naming, I like confirm.sh because it makes it clear it is bash. It doesn't really matter which one you use they function nearly identically, but I would prefer to use the new one `confirm`. I hope its not confusing.

I understood all of that. My assumption was that we _had to_ keep the pre-script-wizard version of `confirm.sh` because we'd need to confirm the _installation_ of script-wizard. We could use `ask` for that but since we want to keep both versions of `confirm`, I was trying to name them so it wouldn't look like there was an old and a new version of the same script and thus one might get accidentally pruned - the name would indicate that one was to be used to confirm s-w installation and the other would be used _with_ s-w after it was installed.

EnigmaCurry:
I'm going to rewrite the `make groups` to use script-wizard

mcmikemn:
I think the comments you added to `confirm` and `confirm.sh` are good enough.

_scripts/reconfigure_auth
```diff
@@ -0,0 +1,57 @@
...
+
+unset_oauth2() {
+    ${BIN}/reconfigure ${ENV_FILE} "${VAR_PREFIX}_OAUTH2"=
+}
+unset_http_auth() {
+    ${BIN}/reconfigure ${ENV_FILE} "${VAR_PREFIX}_HTTP_AUTH"=
+    TMP_PASSWORD=$(mktemp)
+    if [[ -f passwords.json ]]; then
```
mcmikemn, line 22: This check should also me made in `reconfigure_htpasswd`

traefik/README.md
mcmikemn, line 1: there are some lines in here about configuring the traefik dashboard and using HTTP Basic Auth - these should be updated to also mention Oauth2.

traefik/config/setup.sh
```diff
@@ -43,7 +44,7 @@ ytt_template() {
         -v dashboard_entrypoint_enabled="${TRAEFIK_DASHBOARD_ENTRYPOINT_ENABLED}" \
         -v dashboard_entrypoint_host="${TRAEFIK_DASHBOARD_ENTRYPOINT_HOST}" \
         -v dashboard_entrypoint_port="${TRAEFIK_DASHBOARD_ENTRYPOINT_PORT}" \
-        -v dashboard_auth="${TRAEFIK_DASHBOARD_AUTH}" \
+        -v dashboard_auth="${TRAEFIK_DASHBOARD_HTTP_AUTH}" \
```
mcmikemn, line 47: This needs to also be able to handle Oauth2.

traefik/docker-compose.yaml
```diff
@@ -72,7 +73,7 @@ services:
       - TRAEFIK_DASHBOARD_ENTRYPOINT_ENABLED=${TRAEFIK_DASHBOARD_ENTRYPOINT_ENABLED:-false}
       - TRAEFIK_DASHBOARD_ENTRYPOINT_HOST=${TRAEFIK_DASHBOARD_ENTRYPOINT_HOST:-127.0.0.1}
       - TRAEFIK_DASHBOARD_ENTRYPOINT_PORT=${TRAEFIK_DASHBOARD_ENTRYPOINT_PORT:-8080}
-      - TRAEFIK_DASHBOARD_AUTH
+      - TRAEFIK_DASHBOARD_HTTP_AUTH
```
mcmikemn, line 76: I'm pretty sure that `TRAEFIK_DASHBOARD_OAUTH2` needs to be included here, too

mcmikemn: nope

traefik/config/setup.sh
mcmikemn: nope

traefik/README.md
mcmikemn: nope

mcmikemn (commented):
At this point, I think all that's missing is the re-write of `reconfigure_header_authorization_groups` to work with script-wizard.

mcmikemn (commented):
At this point, I think all that's missing is the re-write of `reconfigure_header_authorization_groups` to work with script-wizard.

EnigmaCurry:
`make groups` is now rewritten for script-wizard, I'm sure theres bugs, bash is still not pretty even with the helper, please test for bugs. 

make sure to grab the latest script-wizard by re-running `make install-script-wizard` in the root dir.

mcmikemn:
Great improvement, @EnigmaCurry!

_scripts/reconfigure_header_authorization_groups
```diff
@@ -0,0 +1,250 @@
...
+}
+
+add_group_members() {
+    local GROUP=$1
+    local TRAEFIK_HEADER_AUTHORIZATION_GROUPS=$(${BIN}/dotenv -f ${ENV_FILE} get TRAEFIK_HEADER_AUTHORIZATION_GROUPS)
+    readarray -t users < <(echo "${TRAEFIK_HEADER_AUTHORIZATION_GROUPS}" | jq -r ".\"${GROUP}\"[]")
+    echo "Enter the new user id(s) to add, one per line:"
+    while :
+    do
+        user_id=$(wizard ask "Enter a user ID (press Esc or enter a blank value to finish)" --allow-blank)
```
mcmikemn, line 123: `--allow-blanks` isn't working. 
```
> Security Group Manager (main menu): Group Manager
> Choose a group to manage imagegenerators
> Managing group imagegenerators :: select a task :: Add members
Enter the new user id(s) to add, one per line:
error: unexpected argument '--allow-blank' found

tip: to pass '--allow-blank' as a value, use '-- --allow-blank'

Usage: script-wizard ask <QUESTION|DEFAULT|--json>

For more information, try '--help'.
Wrote new TRAEFIK_HEADER_AUTHORIZATION_GROUPS to ../traefik/.env_server01.thewooskeys.com_default
```

_scripts/reconfigure_header_authorization_groups
```diff
@@ -0,0 +1,250 @@
...
+        echo "Group ${GROUP} already exists."
+        return 0
+    fi
+
+    users=()
+    if wizard confirm "Do you want to add users to this group now?" no; then
+        echo "Enter the new user id(s) to add, one per line:"
+        while :
+        do
+            user_id=$(wizard ask "Enter a user ID (Press Esc or enter a blank value to finish)" --allow-blank)
```
mcmikemn, line 182: `--allow-blanks` isn't working. 
```
> Security Group Manager (main menu): Group Manager
> Choose a group to manage imagegenerators
> Managing group imagegenerators :: select a task :: Add members
Enter the new user id(s) to add, one per line:
error: unexpected argument '--allow-blank' found

tip: to pass '--allow-blank' as a value, use '-- --allow-blank'

Usage: script-wizard ask <QUESTION|DEFAULT|--json>

For more information, try '--help'.
Wrote new TRAEFIK_HEADER_AUTHORIZATION_GROUPS to ../traefik/.env_server01.thewooskeys.com_default
```

_scripts/reconfigure_header_authorization_groups
mcmikemn: fixed with version 0.1.19 of script-wizard

_scripts/reconfigure_header_authorization_groups
mcmikemn: fixed with version 0.1.19 of script-wizard

EnigmaCurry:
I've tested all the apps you configured, they work great!

There's still apps though that aren't configured, and are broken because they can't call `reconfigure_htpasswd`.

EnigmaCurry:
I have used `true` in most of the other .env files so I am changing all the OAUTH2=yes to OAUTH2=true.

EnigmaCurry:
we should have a utility that outputs all the urls you are expected to paste into your gitea app config. it could scan for all the running apps that have the middleware attached and print the list of all of them, then you just copy and paste to gitea.

EnigmaCurry:
I started going alphabetically through the ones I think we still need to fix, I got to nextcloud before I stopped.

mcmikemn:
I tried to instance-iate nextcloud but couldn't make it work. Then I switched to the master branch and simply tried to install it, and I can't make _that_ work.

EnigmaCurry:
I'm gonna skip nextcloud and only focus on the ones that currently have basic auth

EnigmaCurry:
@mcmikemn  I have gone through all the apps that had http basic auth, and upgraded them to use reconfigure_auth. 

I fixed traefik-forward-auth to use auth host mode, so when you create the Gitea oauth application, you only need to enter a single redirect uri : `https://auth.example.com/_oauth`. 

`make groups` is renamed to `make sentry`. This includes a new method to print all the redirect uris, but this shouldn't be necessary as long as auth host mode is working.

Please test creating a new gitea oauth app, reconfigure/redeploy traefik-forward-auth, and then retest a handful of these applications, and carefully read all the options, and see if it a) works and b) makes sense.

EnigmaCurry:
I upgraded script-wizard again, so run `make install-script-wizard` again in the root project dir to upgrade to v0.1.20.

vaultwarden/README.md
```diff
@@ -41,7 +45,7 @@ registration:
 
 ```
 # Allow public registrations, but don't forget to disable this again later.
-make enable-registration
+gmake enable-registration
```
mcmikemn, line 48: Is this supposed to be `gmake` or `make`?

EnigmaCurry: make :+1:

EnigmaCurry:
I removed the `make install-script-wizard` target, and now it checks the ${ROOT_DIR}/.tools.lock.json file for the required locked version instead, and if the version is not installed, it runs the installer after confirmation.

mcmikemn (approved):

PR #122, opened 2023-10-17
Install script-wizard

EnigmaCurry:
This installs [script-wizard](https://github.com/enigmacurry/script-wizard) when you run the root `make config`.

script-wizard will help us make a better UI for the questions that the makefiles ask.

This PR refactors the `confirm` script to use script-wizard.

I have to think more if I want to overwrite the `ask` functions, I don't think theres much to gain by replacing them.

The main benefit of this will be the additional methods of `choose` and `select` which allow choosing from multiple choices.

So far in d.rymcg.tech we have only had binary choices in the makefiles, but now with script wizard we can make multiple choice selections.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ docker-compose.override*.yaml
 dump.*
 tasks.org
 node_modules
+_scripts/script-wizard
--- a/Makefile
+++ b/Makefile
@@ -8,16 +8,24 @@ help:
 include _scripts/Makefile.globals
 include _scripts/Makefile.cd
 
+.PHONY: install-script-wizard
+install-script-wizard:
+	_scripts/install_script-wizard
+
+.PHONY: script-wizard
+script-wizard:
+	@test -f _scripts/script-wizard && (echo "Found script-wizard: $$(_scripts/script-wizard --version) (${ROOT_DIR}/_scripts/script-wizard)" || _scripts/install_script-wizard)
+
 .PHONY: check-deps
 check-deps:
-	_scripts/check_deps docker sed awk xargs openssl htpasswd jq
+	_scripts/check_deps docker sed awk xargs openssl htpasswd jq curl
 
 .PHONY: check-docker
 check-docker:
 	@docker info >/dev/null && echo "Docker is running." || (echo "Could not connect to Docker!" && false)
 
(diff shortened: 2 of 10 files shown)


mcmikemn:
@EnigmaCurry, is `@test -f _scripts/script-wizard &&...` supposed to be `@test -f _scripts/install-script-wizard &&...`, or is `_scripts/script-wizard` missing?
https://github.com/EnigmaCurry/d.rymcg.tech/blob/fbc04f8ad1bf333384f6c5b326ffa7720d19288d/Makefile#L17

EnigmaCurry:
You need to run the root `make config` and it will install script-wizard if it does not exist.

EnigmaCurry:
I added the `reconfigure_choose` script : 

```
## reconfigure_choose ENV_FILE VAR PROMPT CHOICE...
_scripts/reconfigure_choose .env THING "choose a THING" "Thing 1" "Thing 2" "Something else" "None of the above"
```

mcmikemn:
> You need to run the root `make config` and it will install script-wizard if it does not exist.

```
d.rymcg.tech $ make config
make: *** [Makefile:17: script-wizard] Error 1
```

Line 17 seems to reference a file that doesn't exist.

EnigmaCurry:
thanks that should be fixed now.

EnigmaCurry:
I need to make a default value for choose..

EnigmaCurry:
I think its working now, it loads the value and provides the default selection from .env and saves it back now.
```
${BIN}/reconfigure_choose .env THING "Choose a thing" "thing one" "thing two" "thing three"
${BIN}/reconfigure_select .env THINGS "Choose many things" "thing one" "thing two" "thing three"
```
You will need to install the new version of script-wizard by running `make install-script-wizard` in the root.

EnigmaCurry:
Theres another that I don't know if we'll ever use, but you could technically store an entire config BLOB in a single env var:

Launch a full text editor and store what the user types as a JSON string:
```
_scripts/reconfigure_editor .env BIOGRAPHY "Tell me alllll about yourself"
```

EnigmaCurry:
I have merged the script-wizard branch into the header-authorization branch #120, and closing this.

PR #127, opened 2023-10-21
Fix qbit vars

mcmikemn:
I aligned the vars between `.env-dist`, `docker-compose.yaml`, and `qBittorrent.conf`. I installed on fresh droplet and was able to access UI and DL a torrent. Now that this is cleaned and confirmed working, I want to merge it to header-authorization and test it with the new instantiation and auth.


EnigmaCurry:
Thanks, I merged this into #120

PR #131, opened 2023-10-25
Mumble

EnigmaCurry:


EnigmaCurry:
So far this installs mumble and sets up the traefik entrypoint for TCP only. (this will enable the use of the ip whitelist middleware later.)

I would like to be able to automatically configure the ACLs, create channels, and user accounts. As it is, mumble is relatively unsecured right now and is open to all registrations and public access to channels.

PR #133, opened 2023-10-26
plugged in standard shell target to all apps

### LINKED ISSUE #28: Standardize `make shell`
Remove all of the of `make shell` targets in all Makefiles.

Add a new standard shell target in the Makefile.projects like:

```
.PHONY: shell
shell:
	@docker-compose --env-file ${ENV_FILE} exec -it $${service:-frontend} /bin/sh -c "/bin/bash || /bin/sh"
```

The above example needs to properly account for instances too (see #110).

Instead of `frontend` as the default though, we can grep the docker-compose.yaml services and find the first service not named `config` and use that as the default instead. The user can specify their own service name to override this.

If a project needs a specialized shell target, they should be able to specify their own `shell-hook` target that will get run instead of the standard one.

mcmikemn:
I plugged in the new standardized `make shell` target to all apps. I added a call to script-wizard for those that had multiple services one might want to shell to.

Resolves #28

### DIFF
--- a/_attic/cryptpad/Makefile
+++ b/_attic/cryptpad/Makefile
@@ -12,4 +12,4 @@ config-hook:
 
 .PHONY: shell
 shell:
-	make --no-print-directory docker-compose-lifecycle-cmd EXTRA_ARGS="exec -it cryptpad bash"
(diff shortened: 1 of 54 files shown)


EnigmaCurry (approved):
Fantastic

PR #142, opened 2023-10-31
colorize current context in `d.rymcg.tech context`

mcmikemn:
Changed `d.rymcg.tech context` to display the current context in green so it stands out better.

### DIFF
--- a/_scripts/colorize
+++ b/_scripts/colorize
@@ -1,0 +1,1 @@
+run_func
--- a/_scripts/d.rymcg.tech
+++ b/_scripts/d.rymcg.tech
@@ -178,23 +178,7 @@ __info() {
 __context() {
     if [[ $# == 0 ]]; then
         # Run `docker context ls` and store the output in a variable
-        output=$(docker context ls)
-        
-        # Set the color variables
-        green=$(tput setaf 2)      # Green text color
-        reset=$(tput sgr0)        # Reset text formatting
-        
-        # Loop through each line of the output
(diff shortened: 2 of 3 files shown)


EnigmaCurry:
This is a nice idea. Recently I added the [color](https://github.com/EnigmaCurry/d.rymcg.tech/blob/2594f8e30a1db3f8e6e7f7c6e5aa9bd80a813876/_scripts/funcs.sh#L197-L230) function and I would like to consolidate and use that for all text coloring. 

So I think this can be replaced with:

```
        output=$(docker context ls)
        color "light green" "${output}"
```

EnigmaCurry:
you'd still need the code you're doing to print only the one line out of the output

EnigmaCurry:
this might be more efficient

```
d.rymcg.tech context | GREP_COLORS="mt=01;32" grep --color -E '^.* \* .*$|'
```

grep matches for * and greedily expands to that whole line and colors it green.

I want to see if I can generalize that into a reusable function

EnigmaCurry:
LOL you can make it blink 

```
d.rymcg.tech context | GREP_COLORS="ms=05;32" grep --color -E '^.* \* .*$|'
```

EnigmaCurry:
I added a similar `colorize` that reads from stdin and highlights matches.

```
d.rymcg.tech context | colorize green ' \* '
```

wdyt?

mcmikemn:
Awesome.

mcmikemn:
But no blinking. :)

PR #154, opened 2023-11-05
Menu configuration

EnigmaCurry:
This refactors the Traefik `make config` to use a menu driven system made with script-wizard.

https://asciinema.org/a/tGVn2mc8mPNPKIk5wKKHCyjxX

PR #157, opened 2023-11-07
Check all vars from .env-dist are in instance .env file before config.

### LINKED ISSUE #155: `make config` in any project should check the .env file against the .env-dist for missing variables.
Sometimes we add new variables to .env-dist, and this will make it so that existing .env files no longer work with `make config`:

```
> Select an error page theme (https://github.com/tarampampam/error-pages#-templates) l7-light
ERROR: TRAEFIK_ERROR_PAGES_TEMPLATE not found in .env_ssh.d.thewooskeys.com_default
make[1]: *** [Makefile:29: config-ask] Error 1
make: *** [../_scripts/Makefile.lifecycle:4: config] Error 2
```

We should add a pre-check before `config-hook` runs that compares the env vars in .env to .env-dist.

It would be nice if the check would list the missing and/or unused vars and offer to add/remove them in the .env

EnigmaCurry:
Also remove old vars no longer in env-dist.

Fixes #155

### DIFF
--- a/_scripts/Makefile.globals
+++ b/_scripts/Makefile.globals
@@ -48,3 +48,7 @@ check-in-subshell:
 .PHONY: script-wizard
 script-wizard:
 	${BIN}/install_script-wizard
+
+.PHONY: check-dist-vars
+check-dist-vars:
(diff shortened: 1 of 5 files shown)


EnigmaCurry:
I'm closing this and merging into #154 because its easier that way.

PR #161, opened 2023-11-21
Make xbs config template

EnigmaCurry:
This doesn't work yet, when I query https://${XBS_TRAEFIK_HOST}/info I don't see the correct information including the message and location. Why doesn't the setttings.json get read?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 .env
 .env_*
-/xbs/api/settings.json
 /*.tgz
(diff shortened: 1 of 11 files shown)


mcmikemn (approved):

EnigmaCurry:
Thanks mike

mcmikemn:
You're welcome, or rather, my pleasure. And thank YOU!

PR #162, opened 2023-11-22
Adds `d.rymcg.tech git` subcommand

### LINKED ISSUE #114: make `d.rymcg.tech git` subcommand
Because its inconvenient to type `cd ~/git/vendor/enigmacurry/d.rymcg.tech` it would be nice to be able to run git commands directly without needing to change directories.

the git subcommand would just defer to the actual git command, and should also add the tab completion of the git command.

EnigmaCurry:
However, it has no tab completion :( so I don't know how useful this is ...

Sorta fixes #114

### DIFF
--- a/_scripts/d.rymcg.tech
+++ b/_scripts/d.rymcg.tech
@@ -43,6 +43,7 @@ __help() {
         echo -e "make [PROJECT] [ARGS ...]\tRun a \`make\` command for the given d.rymcg.tech project name"
         echo -e "context\tView or set the current docker context"
         echo -e "completion\tSetup TAB completion in your shell"
+        echo -e "git\tRun git commands on the d.rymcg.tech ROOT_DIR"
     ) | expand -t 30
     echo ""
     echo "## Documentation sub-commands:"
@@ -193,6 +194,10 @@ __context() {
     fi
 }
 
+__git() {
+    git -C ${ROOT_DIR} "$@"
+}
+
 __run_script() {
     CMD="${1}"; shift
     ${ROOT_DIR}/_scripts/${CMD} "$@"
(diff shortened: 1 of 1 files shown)


EnigmaCurry:
There is a partial answer to the tab completion question here:

https://stackoverflow.com/a/55668486

However, it only works one sub-command in, it won't complete the whole command list.

Another hint is to check out the definition of the sudo completion, because `sudo git ...` completes a lot more.

Here's what I have the definition of `_sudo` for:

```
_sudo () 
{ 
    local cur prev words cword split;
    _init_completion -s || return;
    local i mode=normal;
    [[ $1 == *sudoedit ]] && mode=edit;
    [[ $mode == normal ]] && for ((i = 1; i <= cword; i++))
    do
        if [[ ${words[i]} != -* ]]; then
            local PATH=$PATH:/sbin:/usr/sbin:/usr/local/sbin;
            local root_command=${words[i]};
            _command_offset $i;
            return;
        fi;
        if [[ ${words[i]} == -@(!(-*)e*|-edit) ]]; then
            mode=edit;
            break;
        fi;
        [[ ${words[i]} == -@(user|other-user|group|close-from|prompt|!(-*)[uUgCp]) ]] && ((i++));
    done;
    case "$prev" in 
        --user | --other-user | -!(-*)[uU])
            COMPREPLY=($(compgen -u -- "$cur"));
            return
        ;;
        --group | -!(-*)g)
            COMPREPLY=($(compgen -g -- "$cur"));
            return
        ;;
        --close-from | --prompt | -!(-*)[Cp])
            return
        ;;
    esac;
    $split && return;
    if [[ $cur == -* ]]; then
        local opts=$(_parse_help "$1");
        COMPREPLY=($(compgen -W '${opts:-$(_parse_usage "$1")}' -- "$cur"));
        [[ ${COMPREPLY-} == *= ]] && compopt -o nospace;
        return;
    fi;
    if [[ $mode == edit ]]; then
        _filedir;
    fi
}
```

so maybe the key is found within, but I haven't figured it out.

PR #180, opened 2024-01-30
Makefile Improvements: English Corrections, Comments Formatting, and …

ZarTek-Creole:
….PHONY Targets

This pull request aims to enhance the Makefile by making the following improvements:

- Corrected English grammar and language for better readability and understanding.
- Formatted comments in a way that is compatible with the `help` target.
- Added `.PHONY` targets to ensure proper execution.

These changes make the Makefile more user-friendly and maintainable. The goal is to provide clear and concise instructions for each target and improve overall code quality.

PR #186, opened 2024-02-04
chore(Makefile): update target for 'userns-remap'

ZarTek-Creole:
feat(Makefile): add support for 'userns-remap on' and 'userns-remap off' commands to configure Docker server for User/Root Namespace Remap
fix(_scripts/userns-remap): update command argument handling to include descriptions for 'on', 'off', and 'check' options

PR for https://github.com/EnigmaCurry/d.rymcg.tech/pull/180#issuecomment-1916596013

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,17 +80,9 @@ audit:
 netstat:
 	ssh $$(docker context inspect $$(${BIN}/docker_context) --format "{{ .Endpoints.docker.Host }}") netstat -plunt
 
-.PHONY: userns-remap # Configure the Docker server for User Namespace Remap
+.PHONY: userns-remap # Configure the Docker server for User/Root Namespace Remap
 userns-remap:
-	@${BIN}/userns-remap true
-
-.PHONY: userns-remap-off # Configure the Docker server for Root Namespace
-userns-remap-off:
-	@${BIN}/userns-remap false
-
-.PHONY: userns-remap-check # Check the current Docker User Namespace Remap setting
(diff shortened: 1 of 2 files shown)


ZarTek-Creole:
for https://github.com/EnigmaCurry/d.rymcg.tech/pull/180#issuecomment-1916596013

EnigmaCurry:
Wow. I've never seen `$(filter-out $@,$(MAKECMDGOALS))` before. I never knew you could make an argument to make like that.

Previously, I was under the impression that:
 * Single words like `config` and `install` are make TARGETS.
 * You can combine several words to chain commands, eg. `make config install`.
 * Make arguments require an `=` symbol. like `make install service=frontend`.

Adding extra non target, non `=` args would break this idea.

What would `make userns-remap off config install` do? In my previous thinking `off` would be its own target.

I have to think more about this. It would be fine for `userns-remap` because its not likely you would want to chanin commands. But if we do it here, we might do it elsewhere, and so it might break a lot of assumptions I was making. Thanks for bringing this technique to my attention.

EnigmaCurry:
The cleaner solution to be compatible with my prior thinking would be something like `make userns-remap enabled=true` and if you just did `make userns-remap` it would ask you interactively should it be on or off.

EnigmaCurry:
I'm inclined to close this for now as I don't want to change the previous convention for named arguments. Thanks again for showing me this cool trick though :+1:

PR #196, opened 2024-05-06
Wireguard (2)

EnigmaCurry:
This is another wireguard config, generic, and separate from the one integrated with Traefik.


EnigmaCurry:
I think it makes sense to re-integrate this idea back into the traefik config, but now with three options:

1)  run traefik as wireguard client (public layer 7 gateway)
2)  run traefik as wireguard server (private layer 7 cloud intranet)
3)  run traefik as TCP only (layer 4) gateway server (private client can connect to IT to expose it as a public server)

EnigmaCurry:
This is still useful for at least the private browsing use-case, so I'm reopening this. Still, a layer 4 traefik proxy option would be useful which would be separate from this.

PR #225, opened 2024-05-28
d.rymcg.tech cli commands: ssh, new-context, install-docker

EnigmaCurry:
Adds automated way of creating the SSH config and Docker context on your workstation.


mcmikemn (approved):
nice work!

PR #235, opened 2024-06-12
Misc fixes

mcmikemn:
Just a few minor fixes I found, and I also wanted `port_forward` to echo the ssh command that creates the tunnel

### DIFF
--- a/_scripts/port_forward
+++ b/_scripts/port_forward
@@ -28,9 +28,8 @@ check_num LOCAL_PORT REMOTE_PORT
 SSH_HOST="$(docker context inspect | jq -r '.[0]["Endpoints"]["docker"]["Host"]' | sed 's|^ssh://||')"
 check_var SSH_HOST
 
-nohup ssh -N -L ${LOCAL_PORT}:${REMOTE_IP_ADDRESS}:${REMOTE_PORT} ${SSH_HOST} &
+${BIN}/exe nohup ssh -N -L ${LOCAL_PORT}:${REMOTE_IP_ADDRESS}:${REMOTE_PORT} ${SSH_HOST} &
 echo "Starting SSH Tunnel ..." >/dev/stderr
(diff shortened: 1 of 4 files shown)


_scripts/port_forward
```diff
@@ -30,6 +30,7 @@ check_var SSH_HOST
 
 nohup ssh -N -L ${LOCAL_PORT}:${REMOTE_IP_ADDRESS}:${REMOTE_PORT} ${SSH_HOST} &
 echo "Starting SSH Tunnel ..." >/dev/stderr
+echo "ssh -N -L ${LOCAL_PORT}:${REMOTE_IP_ADDRESS}:${REMOTE_PORT} ${SSH_HOST}" >/dev/stderr
```
EnigmaCurry, line 33: I don't like that this repeats the options. Theres a few alternatives I would consider instead:

turn on logging temporarily just for the ssh command:

```
set -x
nohup ssh ....
set +x
```

Do the same but from a subshell so you don't need to turn it off. (But does it work with nohup?)

```
(set -x; nohup ssh ....)
```

Setting a variable with the options and referencing it twice:

```
SSH_ARGS="ssh -N -L ${LOCAL_PORT}:${REMOTE_IP_ADDRESS}:${REMOTE_PORT} ${SSH_HOST}"
nohup ${SSH_ARGS}
echo "${SSH_ARGS}
```

EnigmaCurry (changes requested):

_scripts/port_forward
EnigmaCurry: This is also what I made the `exe` alias for: https://github.com/EnigmaCurry/d.rymcg.tech/blob/817fed20ed7a2c8ad5e88eddf114a5c91e1538bd/_scripts/funcs.sh#L10C1-L10C26

So another option might be:

```
exe nohup ssh
```

mcmikemn: Fixed in [6017184](https://github.com/EnigmaCurry/d.rymcg.tech/pull/235/commits/6017184ca037bfbce2670f200bb4a9d3ccb82b21), but I used ticks and single quotes in the commit message which screwed the message up.

EnigmaCurry (approved):

PR #237, opened 2024-06-14
Wireguard Layer 7 TLS and Layer 4 TCP/UDP Proxy

### LINKED ISSUE #236: Refactor Traefik Wireguard Layer 7 proxy
The traefik wireguard setup we have now is hard to configure because the client is set to terminate TLS, so it has two domains for everything, and two sets of TLS certs.

This can be simplified by using a single domain, single certificate, TLS passthrough, and Host-SNI routing.

EnigmaCurry:
Fixes #236


EnigmaCurry:
TODO: 
Ask question to pass proxy protocol on layer 4 from sentry.
Figure out what corrupts layer 7 routes and add spaces to old entries when you add new ones.

EnigmaCurry:
we should test this with the new version of traefik 3.1.0 https://github.com/traefik/traefik/releases/tag/v3.1.0

mcmikemn:
> we should test this with the new version of traefik 3.1.0 https://github.com/traefik/traefik/releases/tag/v3.1.0

@EnigmaCurry, I installed traefik 3.1.0 and tested layer 7 and layer 4 routes: everything works and I saw nothing bad in the logs (except for some "ForceSlash option is deprecated" alerts for tiddlywiki-nodejs, but that's an attic-ed app that may or may not get updated in d.ry, and it still works despite the deprecated call).

mcmikemn (dismissed):
This looks good to me, @EnigmaCurry.

EnigmaCurry:
Thanks for your help @mcmikemn !

PR #242, opened 2024-06-16
add mtls to ttrss

mcmikemn:
Added mTLS to TTRSS and it works fine if you use TTRSS from its web UI. But if you use a client app (e.g., the official [Tiny Tiny RSS for Android](https://gitlab.tt-rss.org/tt-rss/tt-rss-android), or third party Android apps like TTRSS-Reader, Tiny Tiny Feed, Geekttrss), the client app can't connect to the TTRSS Server.

TTRSS does [support SSL cert auth](https://tt-rss.org/wiki/SSL%20Certificate%20Authentication), but _if_ that's going to allow client apps to access the TTRSS server via mTLS, installation of TTRSS might need to be something like:

1. `d make step-ca cert` to create the cert you want your android app to use
2.  `d make ttrss config` would ask which cert in `step-ca/certs` you want to use
3. `d make ttrss install` would build container with `Dockerfile`:
    1. copying the cert to correct directory in web-nginx container
    2. running `sudo update-ca-certificates` in web-nginx container
    3. modifying nginx.conf in web-nginx container
    4. restarting nginx in web-nginx container
5. `d make ttrss install`  would shred the selected cert files (or maybe ask user first)
6. `d make ttrss open` and in TTRSS UI, go to Preferences and scroll to the bottom; under "Login with an SSL certificate" the "Register" button should now be available - click it, then Save configuration.

This PR is probably worth merging as-is, in case someone wants to use TTRSS strictly from its web UI and wants mTLS. But I don't think it's worth the effort of making mTLS work with TTRSS client apps. The official app has support for HTTP Basic Auth.

### DIFF
--- a/ttrss/.env-dist
+++ b/ttrss/.env-dist
@@ -34,4 +34,4 @@ TTRSS_OAUTH2_AUTHORIZED_GROUP=
 TTRSS_MTLS_AUTH=false
 # Enter a comma separated list of client domains allowed to connect via mTLS.
 # Wildcards are allowed and encouraged on a per-app basis:
-TTRSS_MTLS_AUTHORIZED_CERTS=*.clients.whoami.example.com
+TTRSS_MTLS_AUTHORIZED_CERTS=*.clients.ttrss.example.com
--- a/ttrss/.env-dist
+++ b/ttrss/.env-dist
@@ -28,3 +28,10 @@ TTRSS_OAUTH2=
 # by entering which authorization group can log into your app. You create
 # groups of email addresses in the `traefik` folder by running `make groups`.
 TTRSS_OAUTH2_AUTHORIZED_GROUP=
+
+# Mutual TLS (mTLS):
+# Set true or false. If true, all clients must present a certificate signed by Step-CA:
+TTRSS_MTLS_AUTH=false
+# Enter a comma separated list of client domains allowed to connect via mTLS.
+# Wildcards are allowed and encouraged on a per-app basis:
+TTRSS_MTLS_AUTHORIZED_CERTS=*.clients.whoami.example.com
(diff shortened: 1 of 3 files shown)


ttrss/.env-dist
```diff
@@ -28,3 +28,10 @@ TTRSS_OAUTH2=
 # by entering which authorization group can log into your app. You create
 # groups of email addresses in the `traefik` folder by running `make groups`.
 TTRSS_OAUTH2_AUTHORIZED_GROUP=
+
+# Mutual TLS (mTLS):
+# Set true or false. If true, all clients must present a certificate signed by Step-CA:
+TTRSS_MTLS_AUTH=false
+# Enter a comma separated list of client domains allowed to connect via mTLS.
+# Wildcards are allowed and encouraged on a per-app basis:
+TTRSS_MTLS_AUTHORIZED_CERTS=*.clients.whoami.example.com
```
EnigmaCurry, line 37: change whoami to ttrss

EnigmaCurry (dismissed):

EnigmaCurry:
I dont understand why steps 2,3,4 need to happen. You should just need to do step 1 and in config allow the cert name.

EnigmaCurry:
Hmm it sounded like it wants to register the cert in the webapp itself? That wont work then, because its not transparent. You need to be able to register the cert in the client only, without needing to configure ttrss server.

Anyway, it looks good the way it is.

mcmikemn:
> I dont understand why steps 2,3,4 need to happen. You should just need to do step 1 and in config allow the cert name.

You're most likely correct. Step-ca and mTLS is still a bit confusing to me.

I just realized that the TTRSS app itself may not need to know anything about the mTLS certs. I might be able to install the .p12 from step 1 into Android and make sure `TTRSS_MTLS_AUTHORIZED_CERTS=<client cert for my laptop>,<client cert for my android>`, and that might be all it takes.

ttrss/.env-dist
mcmikemn: Fixed in [8d025b3](https://github.com/EnigmaCurry/d.rymcg.tech/pull/242/commits/8d025b3a3b5c5f48cc0d1507d689211485053c53)

EnigmaCurry (approved):

EnigmaCurry:
Thats what I'm thinking, this should be all you need. If not open another issue.

PR #247, opened 2024-06-29
Updated Nextcloud

mcmikemn:
- added instance
- added auth
- added custom mountpoint option
- fixed commented-env-vars issue
- added `make config` options to choose and configure S3 or docker volume for primary storage
- removed erroneous duplicate `restart` in docker-compose.yaml

### DIFF
--- a/nextcloud/docker-compose.instance.yaml
+++ b/nextcloud/docker-compose.instance.yaml
@@ -1,0 +1,91 @@
+#! This is a ytt template file for docker-compose.override.yaml
+#! References:
+#!   https://carvel.dev/ytt
+#!   https://docs.docker.com/compose/extends/#adding-and-overriding-configuration
+#!   https://github.com/enigmacurry/d.rymcg.tech#overriding-docker-composeyaml-per-instance
+
+#! ### Standard project vars:
+#@ load("@ytt:data", "data")
+#@ project = data.values.project
+#@ instance = data.values.instance
+#@ context = data.values.context
+#@ traefik_host = data.values.traefik_host
+#@ ip_sourcerange = data.values.ip_sourcerange
+#@ enable_http_auth = len(data.values.http_auth.strip()) > 0
+#@ http_auth = data.values.http_auth_var
+#@ enable_oauth2 = data.values.oauth2 == "true"
+#@ authorized_group = data.values.authorized_group
+#@ enable_mtls_auth = data.values.enable_mtls_auth == "true"
+#@ mtls_authorized_certs = data.values.mtls_authorized_certs
+#@ enabled_middlewares = []
+
+#! ### Custom project vars
+#@ object_storage = data.values.primary_storage == "S3 Bucket"
+#@ objectstore_s3_host = data.values.objectstore_s3_host
+#@ objectstore_s3_bucket = data.values.objectstore_s3_bucket
+#@ objectstore_s3_key = data.values.objectstore_s3_key
+#@ objectstore_s3_secret = data.values.objectstore_s3_secret
+#@ additional_volume = data.values.additional_volume
+
+#@yaml/text-templated-strings
+#@ if additional_volume != "":
+volumes:
+  (@= additional_volume @):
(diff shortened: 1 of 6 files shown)


mcmikemn:
- bumped postgres to 16
- bumped lobaro/restic-backup-docker to 1.3.2

mcmikemn:
It would be nice to have ffmpeg installed on the app container for Nextcloud apps to use, but I don't want to have to build it the container. I considered volume mapping /usr/bin/ffmpeg on the host to /usr/bin/ffmpeg on the container, but assumed it wasn't a wise idea.

EnigmaCurry:
Looks good, if you want ffmpeg I thinking building it in is the right way to go:

https://github.com/EnigmaCurry/d.rymcg.tech/commit/452d8da4825d66bf017958648556e23b6277b637

I didn't push to this branch so if you want, you can cherry pick that commit:

```
git cherry-pick 452d8da4825d66bf017958648556e23b6277b637
```

EnigmaCurry (dismissed):

nextcloud/docker-compose.yaml
```diff
@@ -48,29 +47,17 @@ services:
...
+      - POSTGRES_PASSWORD=${NEXTCLOUD_DATABASE_PASSWORD}
       - REDIS_HOST=redis
-      - OBJECTSTORE_S3_BUCKET
-      - OBJECTSTORE_S3_SSL=true
-      - OBJECTSTORE_S3_USEPATH_STYLE=true
-      - OBJECTSTORE_S3_AUTOCREATE=true
-      - OBJECTSTORE_S3_KEY
-      - OBJECTSTORE_S3_SECRET
-      - OBJECTSTORE_S3_HOST
-      - OBJECTSTORE_S3_PORT=443
```
EnigmaCurry, line 64: Don't these need to be set from the NEXTCLOUD_OBJECTSTORE_S3_* variables?

nextcloud/docker-compose.instance.yaml
```diff
@@ -0,0 +1,91 @@
...
+      #@ if object_storage:
+      - OBJECTSTORE_S3_HOST=(@= objectstore_s3_host @)
+      - OBJECTSTORE_S3_BUCKET=(@= objectstore_s3_bucket @)
+      - OBJECTSTORE_S3_KEY=(@= objectstore_s3_key @)
+      - OBJECTSTORE_S3_SECRET=(@= objectstore_s3_secret @)
+      - OBJECTSTORE_S3_SSL=true
+      - OBJECTSTORE_S3_PORT=443
+      - OBJECTSTORE_S3_USEPATH_STYLE=true
+      - OBJECTSTORE_S3_AUTOCREATE=true
+      #@ end
```
mcmikemn, line 50: @EnigmaCurry, the `OBJECTSTORE_S3_*` env vars are set in `docker-compose.instance.yaml` if `NEXTCLOUD_PRIMARY_STORAGE=S3 Bucket` because if they're in `docker-compose.yaml` but the user doesn't enter values (i.e., doesn't want to use S3 for primary storage), Nextcloud sees that the env vars exist and tries to install primary storage to S3, and will fail because there are no values.

nextcloud/docker-compose.yaml
mcmikemn: @EnigmaCurry, the OBJECTSTORE_S3_* env vars are set in docker-compose.instance.yaml if NEXTCLOUD_PRIMARY_STORAGE=S3 Bucket because if they're in docker-compose.yaml but the user doesn't enter values (i.e., doesn't want to use S3 for primary storage), Nextcloud sees that the env vars exist and tries to install primary storage to S3, and will fail because there are no values.

https://github.com/EnigmaCurry/d.rymcg.tech/pull/247/files#r1660169968

mcmikemn: see comment below - resolve if satisfied.

EnigmaCurry (approved):

Issue #253, opened 2024-07-05
Need ability to customize config outside of repo.


mcmikemn:
I think the only way to configure prometheus with additional servers to scrape is by editing `d.rymg.tech/prometheus/config/prometheus.yml`. But that file is included in the repo. We should make it so the user can add custom config in a way that can be added to `.gitignore`.

PR #259, opened 2024-07-11
Changes to Prometheus

### LINKED ISSUE #254: Add ability to install cadvisor node exporter without prometheus or grafana
It would be nice if you could optionally install cadvisor and/or node exporter, without prometheus or grafana, so you could use d.ry to install them on additional servers.

mcmikemn:
Fixed #254 
Fixing #253

### DIFF
--- a/prometheus/docker-compose.yaml
+++ b/prometheus/docker-compose.yaml
@@ -11,10 +11,16 @@ services:
       context: config
       args:
         PROMETHEUS_CONFIG_YTT_VERSION: ${PROMETHEUS_CONFIG_YTT_VERSION}
+    profiles:
+      - default
+      - alertmanager
     environment:
       - PROMETHEUS_NODE_EXPORTER_ENABLED
+      - PROMETHEUS_NODE_EXPORTER_TARGETS
       - PROMETHEUS_CADVISOR_ENABLED
+      - PROMETHEUS_CADVISOR_TARGETS
       - PROMETHEUS_ALERTMANAGER_ENABLED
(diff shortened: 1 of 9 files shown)


mcmikemn:
@EnigmaCurry, this works insofar as it seems to correctly format the comma-separated list of targets from the .env file into the correct syntax to be templated into prometheus.yml. I can shell into the Prometheus container and `cat /etc/prometheus/prometheus.yml` and the values are there. 

However, Prometheus isn't activating those targets - they're not available in the UI. In the past, I had manually edited `/etc/prometheus/prometheus.yml` to add these targets, and Prometheus recognized them. Now that d.ry is automatically adding them, Prometheus isn't recognizing them. I'm not sure why.

prometheus/docker-compose.yaml
```diff
@@ -11,10 +11,16 @@ services:
       context: config
       args:
         PROMETHEUS_CONFIG_YTT_VERSION: ${PROMETHEUS_CONFIG_YTT_VERSION}
+    profiles:
```
EnigmaCurry, line 14: I think config shouldnt have profiles, because it should always run?

mcmikemn: It looks like only the prometheus and alertmanager services depend on config, so my thinking is that if you want to install only node-exporter and/or cadvisor (without prometheus or grafana), then you don't need config. Is that not a good way to go?

EnigmaCurry: Ok so those dont need any config, makes sense now

PR #260, opened 2024-07-15
Added PeerTube

mcmikemn:
I built this off the wireguard-layer-7 branch.


EnigmaCurry:
Thanks, since this is based on the work of #237 its not possible to review the changeset (at least not on github), so lets get #237 tested and merged, then I'll take a look at this.

peertube/.env-dist
```diff
@@ -0,0 +1,88 @@
...
+# Database / Postgres service configuration
+PEERTUBE_POSTGRES_USER=postgres
+PEERTUBE_POSTGRES_PASSWORD=
+PEERTUBE_DB_SSL=false
+# Default to Postgres service name "postgres" in docker-compose.yml
+
+# PeerTube server configuration
+# If you need more than one IP as trust_proxy
+# pass them as a comma separated array:
+PEERTUBE_TRUST_PROXY=["127.0.0.1", "loopback", "172.18.0.0/16", "10.13.16.1/32"]
```
EnigmaCurry, line 57: we should probably leave this as just 127.0.0.1 (is loopback really necessary its the same right?) 

We can leave a commented out example for 10.13.16.1, but I'm not sure what 172.18.0.0 even is for.

peertube/.env-dist
```diff
@@ -0,0 +1,88 @@
...
+# May be the hostname of your Custom SMTP server
+PEERTUBE_SMTP_HOSTNAME=postfix
+PEERTUBE_SMTP_PORT=25
+PEERTUBE_SMTP_FROM=
+PEERTUBE_SMTP_TLS=false
+PEERTUBE_SMTP_DISABLE_STARTTLS=false
+PEERTUBE_ADMIN_EMAIL=
+
+# Postfix service configuration
+PEERTUBE_POSTFIX_myhostname=
```
EnigmaCurry, line 76: this variable name seems weird, like is it a requirement of peertube to have that be literally `myhostname` and have it lowercase? I think the var name in the .env file should be all uppercase, and it can be translated in the docker-compose.yaml

`PEERTUBE_POSTFIX_HOSTNAME=myhostname` ?

peertube/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+```
+make config
+```
+
+This will ask you to enter the domain name to use.
+It automatically saves your responses into the configuration file
+`.env_{INSTANCE}`.
+
+### Authentication and Authorization
```
EnigmaCurry, line 25: I know its probably important info, but its the same info as all the other apps right? It probably shouldn't be duplicated in every app README, but maybe we could have a separate document that explains this one time.

peertube/docker-compose.yaml
```diff
@@ -0,0 +1,77 @@
+services:
+  peertube:
+    # If you don't want to use the official image and build one from sources:
+    # build:
+    #   context: .
+    #   dockerfile: ./support/docker/production/Dockerfile.bookworm
+    image: chocobozzz/peertube:${PEERTUBE_IMAGE}
```
EnigmaCurry, line 7: I think PEERTUBE_IMAGE should be the whole image name + version.

peertube/docker-compose.yaml
```diff
@@ -0,0 +1,77 @@
...
+      - PEERTUBE_ADMIN_EMAIL=${PEERTUBE_ADMIN_EMAIL}
+      - POSTFIX_myhostname=${PEERTUBE_POSTFIX_myhostname}
+      - OPENDKIM_DOMAINS=${PEERTUBE_OPENDKIM_DOMAINS}
+      - OPENDKIM_RequireSafeKeys=${PEERTUBE_OPENDKIM_RequireSafeKeys}
+      - PEERTUBE_OBJECT_STORAGE_UPLOAD_ACL_PUBLIC="public-read"
+      - PEERTUBE_OBJECT_STORAGE_UPLOAD_ACL_PRIVATE="private"
+      - PEERTUBE_LOG_LEVEL=${PEERTUBE_LOG_LEVEL}
+    ports:
+      # Comment next line if you don't want to use the live feature
+      - "1935:1935"
```
EnigmaCurry, line 32: We don't want to expose any ports, we should create traefik entrypoints if we need additional ports.

peertube/docker-compose.yaml
```diff
@@ -0,0 +1,77 @@
...
+      - data:/data
+      - config:/config
+    depends_on:
+      - postgres
+      - redis
+      - postfix
+    restart: "unless-stopped"
+
+  postgres:
+    image: postgres:${PEERTUBE_POSTGRES_IMAGE}
```
EnigmaCurry, line 45: PEERTUBE_POSTGRES_IMAGE should be the whole image name + version.

peertube/.env-dist
mcmikemn: I don't know what these values should be, though I think 10.13.16.1 is correct for d.ry. The other values were in the source compose.yaml I started from, and I just left them - they can probably be deleted. 

Does this seem good?
```
# PeerTube server configuration
# If you need more than one IP as trust_proxy, pass them as a comma separated array.
# e.g., PEERTUBE_TRUST_PROXY=["127.0.0.1", "10.13.16.1/32"]
PEERTUBE_TRUST_PROXY=["127.0.0.1"]
```

peertube/.env-dist
mcmikemn: fixed in [a3a8e3a](https://github.com/EnigmaCurry/d.rymcg.tech/pull/260/commits/a3a8e3aff9db0591c542a3b2cd8f8b0a6c0a42f1)

peertube/docker-compose.yaml
mcmikemn: fixed in [a3a8e3a](https://github.com/EnigmaCurry/d.rymcg.tech/pull/260/commits/a3a8e3aff9db0591c542a3b2cd8f8b0a6c0a42f1)

peertube/README.md
mcmikemn: I used whoami as a template. until we improve the template project, I'll have to remember to clean the readme better.

fixed in [a3a8e3a](https://github.com/EnigmaCurry/d.rymcg.tech/pull/260/commits/a3a8e3aff9db0591c542a3b2cd8f8b0a6c0a42f1)

peertube/docker-compose.yaml
mcmikemn: I'm going to need help with this one.

peertube/docker-compose.yaml
mcmikemn: fixed in [a3a8e3a](https://github.com/EnigmaCurry/d.rymcg.tech/pull/260/commits/a3a8e3aff9db0591c542a3b2cd8f8b0a6c0a42f1)

peertube/.env-dist
EnigmaCurry: yea

peertube/docker-compose.yaml
EnigmaCurry: I added the entrypoints in #3f4e49a 
You can use the rtmp or rtmps entrypoints.
Example in forgejo where it uses the ssh entrypoint https://github.com/EnigmaCurry/d.rymcg.tech/blob/033bf8e3331b8ca8dcb945380b4365780f855a7a/forgejo/docker-compose.instance.yaml#L33-L37

EnigmaCurry: I didn't add an entrypoint for the 9000 port, because the comment indicated it wasnt for prod use.

EnigmaCurry: README should indicate that to enable RTMP you have to enable the associated traefik entrypoint

EnigmaCurry: For RTMPS you would have traefik terminate the TLS connection with a lets encrypt cert on port 1936 and forward it unencrypted to the peertube service port 1935.

You only need to set a blank tls section and do the port translation:

```
- "traefik.tcp.services.(@= router @)-rtmps.tls=true" 
- "traefik.tcp.services.(@= router @)-rtmps.loadbalancer.server.port=1935"
```

https://doc.traefik.io/traefik/routing/routers/#general_2

peertube/.env-dist
mcmikemn: fixed in [117ae17](https://github.com/EnigmaCurry/d.rymcg.tech/pull/260/commits/117ae17d724b9ae0fc5115e157d9ebcd0fd34a88)

peertube/docker-compose.yaml
mcmikemn: I don't understand this. :(

[25b8edd](https://github.com/EnigmaCurry/d.rymcg.tech/pull/260/commits/25b8edde28b3b8e13d08b55f16300e85b035932e) is my attempt at using other code as a template to guess what should be done here.

peertube/Makefile
```diff
@@ -33,10 +33,12 @@ config-hook:
 		echo && echo "Postfix configuration:" && \
 		${BIN}/reconfigure_ask ${ENV_FILE} PEERTUBE_POSTFIX_myhostname "Enter the domain of your email address" && \
 		# I use `dotenv set` in the next line instead of `reconfigure` because the value needs to have an "=" in it but `reconfigure` strips all "=" \
-		${BIN}/dotenv -f ${ENV_FILE} set PEERTUBE_OPENDKIM_DOMAINS="$$(${BIN}/dotenv -f ${ENV_FILE} get PEERTUBE_POSTFIX_myhostname)=postfix" && echo "Set PEERTUBE_OPENDKIM_DOMAINS=$$(${BIN}/dotenv -f ${ENV_FILE} get PEERTUBE_POSTFIX_myhostname)=postfix" && \
+		${BIN}/dotenv -f ${ENV_FILE} set PEERTUBE_OPENDKIM_DOMAINS="$$(${BIN}/dotenv -f ${ENV_FILE} get PEERTUBE_POSTFIX_myhostname)=postfix" && echo "Set PEERTUBE_OPENDKIM_DOMAINS=$$(${BIN}/dotenv -f ${ENV_FILE} get PEERTUBE_POSTFIX_MYHOSTNAME)=postfix" && \
```
EnigmaCurry, line 36: I still don't understand why this variable has the word "my" in it.

mcmikemn: I started from their example [`docker-compose.yml`](https://github.com/Chocobozzz/PeerTube/blob/master/support/docker/production/docker-compose.yml) and [`.env`](https://github.com/Chocobozzz/PeerTube/blob/master/support/docker/production/.env), and their `.env` has [`POSTFIX_myhostname`](https://github.com/Chocobozzz/PeerTube/blob/3d4d49a23eae71f3ce62cbbd7d93f07336a106b7/support/docker/production/.env#L43) in it.

My assumption was that the postfix container would be looking for an env var called exactly "POSTFIX_myhostname", so I didn't change it.

mcmikemn: I can't tell either. It looks like you should change the var name, but how would the postfix container know to get the value of `POSTFIX_BANANA`?

EnigmaCurry: I guess its a postfix thing, https://unix.stackexchange.com/questions/58243/which-name-should-be-used-for-myhostname-in-postfixs-main-cf

EnigmaCurry: ok whatevs.

peertube/docker-compose.instance.yaml
```diff
@@ -34,6 +37,20 @@ services:
...
+      - "traefik.tcp.routers.(@= router @)-rtmp.rule=HostSNI(`*`)"
+      - "traefik.tcp.routers.(@= router @)-rtmp.entrypoints=rtmp"
+      - "traefik.tcp.routers.(@= router @)-rtmp.service=(@= router @)-rtmp"
+      - "traefik.tcp.services.(@= router @)-rtmp.loadbalancer.server.port=1935"
+      #@ end
+
+      #! RTMPS
+      #@ if livestreaming == "RTMPS":
+      - "traefik.tcp.services.(@= router @)-rtmps.tls=true" 
+      - "traefik.tcp.services.(@= router @)-rtmps.loadbalancer.server.port=1935"
```
EnigmaCurry, line 51: I think its 1936 in the case of rtmps

EnigmaCurry: oh nope, cause we terminate the tls.

peertube/docker-compose.yaml
EnigmaCurry: yea that looks right.

EnigmaCurry: well youd need two routers to handle both rtmp or rtmps, but it seems like any rtmp client is capable of configuring a port, so I think my commit earlier where I made both entrypoints should be consolidated to a single entrypoint `rtmp` and the user can configure if they want TLS or not themselves.

so for rtmps I think it would be:

```
      #! RTMPS
      #@ if livestreaming == "RTMPS":
      - "traefik.tcp.services.(@= router @)-rtmp.tls=true" 
      #@ end
```

EnigmaCurry: I deleted my comment about needing another variable (I know you read email notifications though)

EnigmaCurry: yea

mcmikemn: How does [9d55891](https://github.com/EnigmaCurry/d.rymcg.tech/pull/260/commits/9d5589148474bd799bc310e997392fe7bff66473) look? Peertube works, but I wasn't able to successfully test livestreaming - mainly because I've never livestreamed and don't know how. I started a livestream in Peertube and entered the URL (e.g., rtmp://peertube.thewooskeys.com:1935/live) into VLC, but VLC can't connect. I tried rtp://, http://, and https://. I do have an entrypoint and layer 4 route on the sentry (error free in Traefik dashboard) and an entrypoint on the app server (also error free in Traefik dashboard).

EnigmaCurry: looking good. I would test with OBS.

mcmikemn: Same result: OBS can't connect. So either it doesn't work as it is or, more likely, I don't have it set up correctly.

mcmikemn: `rtmpdump` also can't connect. I think I must have the route configured incorrectly but I don't yet see how.

mcmikemn: My bad. I neglected to open 1935 on the sentry's firewall. Everything works! I think this PR can be merged, @EnigmaCurry.

EnigmaCurry: nice, you tested with RTMP and RTMPS? OBS accepts the certificate?

EnigmaCurry (approved):

PR #262, opened 2024-07-28
Instantiate TFA

mcmikemn:
Adding instances to TFA, mainly so we can define COOKIE_NAME so it differs between TFA on different servers.


mcmikemn:
At the moment, @EnigmaCurry, this only works if you leave COOKIE_NAME to the default value of "_forward_auth" (and you have to make sure that TRAEFIK_FORWARD_AUTH_SECRET is the same value as it is for the other TFA instances since they're all using the same cookie name - otherwise apps using one TFA instance will authenticate and others will fail).

If you set COOKIE_NAME to something unique, like "_forward_auth_context_instance", you get infinite redirects between `app.mydomain.com` and `auth.app.mydomain.com`.

PR #271, opened 2024-09-10
Add Immich

mcmikemn:
This is almost done. It works as long as you use a named Docker volume instead of a bind mount (i.e., if `IMMICH_UPLOAD_LOCATION=media` it works ~, but if `IMMICH_UPLOAD_LOCATION=/media` the container fails to access it). I think this is a permissions issue but I haven't figured it out yet.~

(Also, I haven't tested it yet with auth in front of it - I'll do that once the bind mount is working.)


immich/docker-compose.yaml
```diff
@@ -0,0 +1,65 @@
+services:
+  immich:
+    image: ${IMMICH_IMAGE:-ghcr.io/immich-app/immich-server:release}
+    volumes:
+      - ${IMMICH_UPLOAD_LOCATION}:/usr/src/app/upload
```
EnigmaCurry, line 5: I don't really like this style of env var substitution, it puts too much required knowledge onto the editor of the .env file. 

Instead, I would create a config like "IMMICH_USE_EXTERNAL_VOLUME=true" and "IMMICH_EXTERNAL_VOLUME=/path/to/photos" 

In the template you can discriminate on IMMICH_USE_EXTERNAL_VOLUME and configure the volume this way:

```
version: '3.8'

services:
  my_service:
    image: my_image
    volumes:
      - photos:/app/data

volumes:
  photos:
    driver: local
    driver_opts:
      type: none
      o: bind
      device: /custom/path/on/host
```

that way the named volume is always named photos.

Furthermore, all the other volumes should also be named inside the `volumes` section.

mcmikemn: I re-did the functionality surrounding mapping a bind mount vs. a docker volume ([8f077bd](https://github.com/EnigmaCurry/d.rymcg.tech/pull/271/commits/8f077bdcafbdfd673a868dfcf5132701e91c36f3)).

Now, though, the container is recognizing neither the docker volume _nor_ the bind mount.

immich/.env-dist
```diff
@@ -44,8 +44,13 @@ IMMICH_MTLS_AUTH=false
...
 
-# Enter the path on the host where you want Immich to upload new images (e.g., `/mnt/photos`, `~/images`, `../pictures`), or enter a name to be used as a named Docker volume which will be created for you (e.g., photos).
-IMMICH_UPLOAD_LOCATION=photos
+# Set to "false" to have Immich upload images to a named Docker volume,
+# or to "true" to have Immich upload images to a bind mount directory on
+# the host.
+IMMICH_USE_EXTERNAL_VOLUME=false
+# If IMMICH_USE_EXTERNAL_VOLUME=true, set to the path on the host that you
+# want Immich to upload new images to (e.g., `/mnt/photos` or `~/images`).
+IMMICH_UPLOAD_LOCATION=
```
EnigmaCurry, line 53: If this variable is only used when IMMICH_USE_EXTERNAL_VOLUME is true, then the variable name should reference that. I propose IMMICH_EXTERNAL_VOLUME.

immich/docker-compose.yaml
mcmikemn: My mistake: the container recognizes both docker volumes and bind mounts.

immich/.env-dist
mcmikemn: Fixed in [494f1b8](https://github.com/EnigmaCurry/d.rymcg.tech/pull/271/commits/494f1b8dcd1d9cac97ee4d883baa1f0c2a2282e1)

mcmikemn:
Redis gives this warning:
`WARNING Memory overcommit must be enabled! Without it, a background save or replication may fail under low memory condition. Being disabled, it can can also cause failures without low memory condition, see https://github.com/jemalloc/jemalloc/issues/1328. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect.`

Is this something we should deal with?

mcmikemn:
Before approval, I need to:

- [ ] test Oauth2, mTLS, and basic HTTP auth
- [ ] test how Immich handles bind mounts that already contain images

EnigmaCurry:
> Redis gives this warning: `WARNING Memory overcommit must be enabled! Without it, a background save or replication may fail under low memory condition. Being disabled, it can can also cause failures without low memory condition, see https://github.com/jemalloc/jemalloc/issues/1328. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect.`
> 
> Is this something we should deal with?

It seems to reference a host configuration issue. Since the user may be using widely different hosts, the only thing we can do is to document the potential issue in the README

mcmikemn:
d.ry auth works with immich. I think this is ready to go.

EnigmaCurry (approved):

PR #273, opened 2024-09-11
Removed TLS from reloader_router

mcmikemn:
Removed TLS from reloader_router because webhook would fail if Homepage was installed behind mTLS, and the webhook is protected by a secret key anyway


EnigmaCurry (approved):

EnigmaCurry:
I reverted this in https://github.com/EnigmaCurry/d.rymcg.tech/commit/3319a914be855d7587c2a6b4a1904ceb23f66222 because its not allowed to have different tls options on routers, they need to all be the same.

PR #275, opened 2024-09-12
Fix homepage webhooks TLS options by moving them to a separate domain

EnigmaCurry:
This is so that it can use separate TLS options for the webhooks.


EnigmaCurry:
image

https://doc.traefik.io/traefik/routing/routers/#options

This re-fixes #273 by moving the webhooks to a separate domain.

mcmikemn:
LGTM!

mcmikemn (approved):

PR #277, opened 2024-09-20
Immich updates

mcmikemn:
bump immich version, remove IMMICH_LOG_LEVEL (because you can change the log level from the UI _unless_ it's set during install via an env var).


immich/.env-dist
```diff
@@ -1,7 +1,7 @@
 # The docker image to use for Immich Server (https://github.com/immich-app/immich/pkgs/container/immich-server):
-IMMICH_IMAGE=ghcr.io/immich-app/immich-server:pr-12438 
+IMMICH_IMAGE=ghcr.io/immich-app/immich-server:pr-12772
 # The docker image to use for Immich Machine Learning (https://github.com/immich-app/immich/pkgs/container/immich-machine-learning):
-IMMICH_ML_IMAGE=ghcr.io/immich-app/immich-machine-learning:pr-12456-cuda
+IMMICH_ML_IMAGE=ghcr.io/immich-app/immich-machine-learning:pr-12747-cuda
```
EnigmaCurry, line 4: Is this image stable? Will it disappear when its pr is merged? Perhaps just add some extra comment about why this image is used.

EnigmaCurry (approved):

immich/.env-dist
mcmikemn: @EnigmaCurry, I thought tags (and thus tagged images) were permanent (i.e., they aren't automatically deleted when a PR/branch/etc. is deleted - they are only ever deleted manually). Is that not the case?

EnigmaCurry: I doubt its permanent, but I don't know the specifics of ghcr. They certainly aren't permanent on docker hub.

mcmikemn: I do think they're "permanent" (no automatic deletion), and Immich has 100s of old images still available. But [7cda96b](https://github.com/EnigmaCurry/d.rymcg.tech/pull/277/commits/7cda96b2b0cf1e0784f4538e20a5b6967711b649) pins images tied to a specific release rather than a specific PR.

EnigmaCurry (approved):

PR #279, opened 2024-09-21
Page status results with less -FSX

### LINKED ISSUE #278: Make status should have no line wraps
The status lines are quite long and wrap on my terminal, so it would be more useful if it were piped to `less -FSX` to make it scroll to the right. This has the annoyance now that you have to type Q to quit less (but only if it wraps or is more than one page), but it is far more readable.

d status 

vs 

d status | less -FSX

EnigmaCurry:
Fixes #278


EnigmaCurry:
@mcmikemn WDYT?

mcmikemn:
LGTM. :)

mcmikemn (approved):

PR #280, opened 2024-09-22
Add YOURLS

mcmikemn:
A URL shortener.


mcmikemn:
Fixes #281

yourls/Makefile
```diff
@@ -0,0 +1,106 @@
...
+	done
+# Copy the modified config.php back into the container
+	@docker cp -a config.php.temp yourls-yourls-1:/var/www/html/user/config.php > /dev/null 2>&1
+# Clean up temporary file
+	@rm config.php.temp
+	@echo
+
+.PHONY: add-plugins # Add plugins
+add-plugins:
+	@tar -C ${ROOT_DIR}/yourls/plugins -cf - . | docker exec -i yourls-yourls-1 tar -xf - -C /var/www/html/user/plugins/
```
EnigmaCurry, line 103: I think this should be baked into a Dockerfile so the plugins are part of the image.

EnigmaCurry: I don't think we should be distributing plugins in our repo, can they be downloaded as part of the Dockerfile, or simply a blank directory that the user must populate manually?

mcmikemn: So YOURLS comes with a few plugins installed by default. I wanted a few more installed by default when YOURLS is installed from d.ry, but I also wanted each user to decide for themselves which plugins to add. 

At the moment, I don't have a way to delete a plugin once you install it (you can _deactivate_ it from the UI, but you'd have to actually delete it manually in the container).

Do you still think that the set of plugins that _I_ want as default should be baked into the d.ry image? If so, should I remove the Makefile target to add more plugins, and just let the user do everything manually at that point?

EnigmaCurry: made notes in convo

EnigmaCurry: I think we should get rid of the add-plugins target, and build it into the Dockerfile instead.

EnigmaCurry: like in traefik for example:

https://github.com/EnigmaCurry/d.rymcg.tech/blob/4cd51673e2ae1ef3e7676ba6351b00cb656dbdd1/traefik/config/Dockerfile#L8C1-L8C41

https://github.com/EnigmaCurry/d.rymcg.tech/tree/master/traefik/config/context-template

yourls/yourls/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+		libpng-dev \
+        libzip-dev \
+        unzip \
+        micro \
+    && docker-php-ext-configure zip \
+    && docker-php-ext-install -j$(nproc) zip \
+	&& apt-get clean && rm -rf /var/lib/apt/lists/*
+RUN mkdir -p /var/www/html/user/plugins/yourls-download-plugin && curl -o /var/www/html/user/plugins/yourls-download-plugin/plugin.php https://raw.githubusercontent.com/krissss/yourls-download-plugin/refs/heads/master/plugin.php
+RUN mkdir -p /var/www/html/user/plugins/force-lowercase &&curl -o /var/www/html/user/plugins/force-lowercase/plugin.php https://raw.githubusercontent.com/YOURLS/force-lowercase/refs/heads/master/plugin.php
+RUN mkdir -p /var/www/html/user/plugins/yourls-change-password-master && curl -o /var/www/html/user/plugins/yourls-change-password-master/plugin.php https://raw.githubusercontent.com/ozh/yourls-change-password/refs/heads/master/plugin.php
```
EnigmaCurry, line 15: These URLS should be in the .env file like IMMICH_PLUGIN_CHANGE_PASSWORD_PACKAGE_DIST

yourls/.env-dist
```diff
@@ -0,0 +1,58 @@
...
+YOURLS_OAUTH2_AUTHORIZED_GROUP=
+
+# Mutual TLS (mTLS):
+# Set true or false. If true, all clients must present a certificate signed by Step-CA:
+YOURLS_MTLS_AUTH=false
+# Enter a comma separated list of client domains allowed to connect via mTLS.
+# Wildcards are allowed and encouraged on a per-app basis:
+YOURLS_MTLS_AUTHORIZED_CERTS=*.clients.yourls.example.com
+
+YOURLS_PROXY=10.13.16.1
```
EnigmaCurry, line 43: Please add comments describing what this is.

EnigmaCurry: Does it work if you're not using the VPN?

yourls/Makefile
```diff
@@ -0,0 +1,99 @@
...
+include ${ROOT_DIR}/_scripts/Makefile.projects
+include ${ROOT_DIR}/_scripts/Makefile.instance
+
+.PHONY: config-hook
+config-hook:
+	@${BIN}/reconfigure_ask ${ENV_FILE} YOURLS_TRAEFIK_HOST "Enter the YOURLS domain name" yourls${INSTANCE_URL_SUFFIX}.${ROOT_DOMAIN}
+	@${BIN}/reconfigure ${ENV_FILE} YOURLS_INSTANCE=$${instance:-default}
+	@${BIN}/reconfigure_auth ${ENV_FILE} YOURLS
+	@echo
+	@${BIN}/reconfigure_ask ${ENV_FILE} YOURLS_USER "Enter the admin username for your YOURLS instance"
```
EnigmaCurry, line 11: I leave it up to you, but would you rather type a username and password here, or just have use admin and a random password?

EnigmaCurry: actually, does this use the YOURLS_PASS forever, or does it get copied into the config and then it no longer works if you change it? We dont want env files that cant be changed.

yourls/docker-compose.yaml
```diff
@@ -0,0 +1,44 @@
...
+    depends_on:
+      - mysql
+    environment:
+      YOURLS_DB_PASS: ${MYSQL_ROOT_PASSWORD}
+      YOURLS_SITE: https://${YOURLS_TRAEFIK_HOST}
+      YOURLS_USER: ${YOURLS_USER}
+      YOURLS_PASS: ${YOURLS_PASS}
+      YOURLS_UNIQUE_URLS: ${YOURLS_UNIQUE_URLS}
+      YOURLS_COOKIEKEY: ${YOURLS_COOKIEKEY}
+      YOURLS_USER_PASSWORDS: ${YOURLS_USER_PASSWORDS}
```
EnigmaCurry, line 17: I dont see this var defined anywhere?

yourls/Makefile
mcmikemn: There doesn't appear to be anything special about the "admin" user in YOURLS. You can have multiple users access the "admin" interface. The admin interface lets you have more control over the shortened URLs, as well as control over how YOURLS works in general. Since you might configure multiple users to access the "admin"interface, I think it's fine to let the "admin" username and password be user-defined instead of assigned by d.ry.

yourls/docker-compose.yaml
mcmikemn: It's explained [here](https://yourls.org/docs/guide/essentials/configuration#yourls_cookiekey). 

`make config` assigns a random value:
```
	@${BIN}/reconfigure_password ${ENV_FILE} YOURLS_COOKIEKEY 45
```

yourls/.env-dist
mcmikemn: [This](https://yourls.org/docs/guide/advanced/proxy#proxy-with-authentication) explains what YOURLS_PROXY is for (well, barely). My assumption was that without this var, YOURLS wouldn't function correctly or at all. I haven't tested without YOURLS_PROXY.

yourls/docker-compose.yaml
mcmikemn: fixed in [2dd4f1a](https://github.com/EnigmaCurry/d.rymcg.tech/pull/280/commits/2dd4f1a8ff212abf77e3394c1b825498f9b8b47a)

yourls/yourls/Dockerfile
mcmikemn: fixed in [04d9dbd](https://github.com/EnigmaCurry/d.rymcg.tech/pull/280/commits/04d9dbd64e37a0c2ad041da0000f8184d2476b7e)

yourls/.env-dist
EnigmaCurry: The way I read that, it means to use this as an outgoing proxy, which you don't need.

mcmikemn: Ah, OK. I removed `YOURLS_PROXY` in [de78ef4](https://github.com/EnigmaCurry/d.rymcg.tech/pull/280/commits/de78ef4b69ba5106d900bc0332b3f7e8363bc696)

EnigmaCurry (approved):

PR #284, opened 2024-09-24
Added Glances

mcmikemn:
Added Glances, an open-source system cross-platform monitoring tool. It's like a light version of Prometheus.

I added it because Homepage has a widget for it and I wanted to see basic server stats from Homepage. (Homepage also has a widget for Prometheus, but it's far more basic - it's funny: Promethus is a more robsut metrics monitor but the Homepage widget is basic, Glances is a farily basic metrics monitor but the widget is more robust).

### DIFF
--- a/glances/.env-dist
+++ b/glances/.env-dist
@@ -40,11 +40,19 @@ GLANCES_MTLS_AUTH=false
 # Wildcards are allowed and encouraged on a per-app basis:
 GLANCES_MTLS_AUTHORIZED_CERTS=*.clients.glances.example.com
 
-# Timezone
-GLANCES_TZ=EST5EDT
+# Timezone (see: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones)
+GLANCES_TZ=America/New_York
 
 # Nvidia GPU support
 GLANCES_GPU=false
+
+# Set to "true" to allow Glances access to the host's Docker socket, which will
+# allow Glances to report on metrics for containers. Or set to "false" to
+# disallow access to the host's Docker socket. Be aware that allowing access to
+# the Docker socket is not safe because it effectively grants full control over
+# the Docker daemon, enabling a container or attacker to escalate privileges,
+# manipulate containers, and potentially compromise the host system.
+GLANCES_CONTAINER_METRICS=false
 
 # Glances server runtime options
 # See: https://glances.readthedocs.io/en/develop/cmds.html#command-line-options
--- a/glances/.env-dist
+++ b/glances/.env-dist
@@ -1,0 +1,51 @@
+# The docker image to use:
+GLANCES_IMAGE=nicolargo/glances:4121
+
+# The domain name for the glances service:
+GLANCES_TRAEFIK_HOST=glances.example.com
+
+# The name of this instance. If there is only one instance, use 'default'.
+GLANCES_INSTANCE=
+
+# Filter access by IP address source range (CIDR):
+##Disallow all access: 0.0.0.0/32
+##Allow all access: 0.0.0.0/0
+GLANCES_IP_SOURCERANGE=0.0.0.0/0
+
+# HTTP Basic Authentication:
+# Use `make config` to fill this in properly, or set this to blank to disable.
+GLANCES_HTTP_AUTH=
+
+# OAUTH2
+# Set to `true` to use OpenID/OAuth2 authentication via the
+# traefik-forward-auth service in d.rymcg.tech.
+# Using OpenID/OAuth2 will require login to access your app,
+# but it will not affect what a successfully logged-in person can do in your
+# app. If your app has built-in authentication and can check the user
+# header that traefik-forward-auth sends, then your app can limit what the
+# logged-in person can do in the app. But if your app can't check the user
+# header, or if your app doesn't have built-in authentication at all, then
+# any person with an account on your Gitea server can log into your app and
+# have full access.
+GLANCES_OAUTH2=
+# In addition to Oauth2 authentication, you can configure basic authorization
+# by entering which authorization group can log into your app. You create
+# groups of email addresses in the `traefik` folder by running `make groups`. 
+GLANCES_OAUTH2_AUTHORIZED_GROUP=
+
+# Mutual TLS (mTLS):
+# Set true or false. If true, all clients must present a certificate signed by Step-CA:
+GLANCES_MTLS_AUTH=false
+# Enter a comma separated list of client domains allowed to connect via mTLS.
+# Wildcards are allowed and encouraged on a per-app basis:
+GLANCES_MTLS_AUTHORIZED_CERTS=*.clients.glances.example.com
+
+# Timezone
+GLANCES_TZ=EST5EDT
+
+# Nvidia GPU support
+GLANCES_GPU=false
+
+# Glances server runtime options
+# See: https://glances.readthedocs.io/en/develop/cmds.html#command-line-options
+GLANCES_OPTIONS="-t 2"
--- a/glances/.env-dist
+++ b/glances/.env-dist
@@ -1,5 +1,5 @@
 # The docker image to use:
-GLANCES_IMAGE=nicolargo/glances:4121
+GLANCES_IMAGE=nicolargo/glances:4.1.2.1
 
 # The domain name for the glances service:
 GLANCES_TRAEFIK_HOST=glances.example.com
--- a/glances/docker-compose.instance.yaml
+++ b/glances/docker-compose.instance.yaml
@@ -1,0 +1,79 @@
+#! This is a ytt template file for docker-compose.override.yaml
+#! References:
+#!   https://carvel.dev/ytt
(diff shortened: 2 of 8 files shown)


glances/.env-dist
```diff
@@ -0,0 +1,51 @@
...
+
+# Mutual TLS (mTLS):
+# Set true or false. If true, all clients must present a certificate signed by Step-CA:
+GLANCES_MTLS_AUTH=false
+# Enter a comma separated list of client domains allowed to connect via mTLS.
+# Wildcards are allowed and encouraged on a per-app basis:
+GLANCES_MTLS_AUTHORIZED_CERTS=*.clients.glances.example.com
+
+# Timezone
+GLANCES_TZ=EST5EDT
```
EnigmaCurry, line 44: Does it support timezones like `America/New_York`? I think that would be a more intuitive example.

glances/docker-compose.instance.yaml
```diff
@@ -0,0 +1,79 @@
...
+        #@ enabled_middlewares.append("mtls-header@file")
+      #@ end
+      
+      #! Override the default port that the app binds to:
+      #! You don't normally need to do this, as long as your image has
+      #! an EXPOSE directive in it, Traefik will autodetect it, but this is how you can override it:
+      - "traefik.http.services.(@= router @).loadbalancer.server.port=61208"
+
+      #! Apply all middlewares (do this at the end!)
+      - "traefik.http.routers.(@= router @).middlewares=(@= ','.join(enabled_middlewares) @)"
```
EnigmaCurry, line 79: Glances exposes two services, should there be a second router for the XMLRPC service? 

https://github.com/nicolargo/glances/blob/24c87253595f9d3168afc26e82cfa464fc7523b9/docker-files/ubuntu.Dockerfile#L101

glances/docker-compose.yaml
```diff
@@ -0,0 +1,15 @@
+services:
+  glances:
+    image: ${GLANCES_IMAGE}
+    restart: unless-stopped
+    pid: host
+    privileged: true
```
EnigmaCurry, line 6: We should not use privileged: true, instead follow SECURITY.md and find the minimum capabilities required.

glances/docker-compose.yaml
```diff
@@ -0,0 +1,15 @@
...
+    network_mode: "host"
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock:ro
```
EnigmaCurry, line 9: Is docker required for all functionality?

glances/docker-compose.yaml
```diff
@@ -0,0 +1,15 @@
+services:
+  glances:
+    image: ${GLANCES_IMAGE}
+    restart: unless-stopped
+    pid: host
+    privileged: true
+    network_mode: "host"
```
EnigmaCurry, line 7: Why is network mode host? This is insecure for traefik.

glances/docker-compose.yaml
EnigmaCurry: https://github.com/EnigmaCurry/d.rymcg.tech/blob/master/SECURITY.md#drop-system-capabilities

glances/.env-dist
EnigmaCurry: link in the comments if appropriate: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

glances/docker-compose.instance.yaml
mcmikemn: I don't know. It works for me with just port 61208. Maybe XMLRPC is used for other funtions I'm not using (e.g., Glances has the ability to report on multiple servers - maybe it uses XMLRPC to communicate between Glances instances?)?

`docker ps` displays this container with no ports:
```
<containerId>   nicolargo/glances:4.1.2.1                                "/bin/sh -c '/venv/b…"   2 hours ago    Up 2 hours                                  glances-glances-1
```

glances/docker-compose.yaml
mcmikemn: I think Glances only uses docker socket access (read-only) to report metrics of containers. My personal use-case is for servers, not containers, so I could remove docker socket access. Or we could variablize it and leave it up to the user?

EnigmaCurry: Unfortunately the `ro` is completely and total LIE. This gives read and write privileges.

glances/docker-compose.yaml
mcmikemn: ChatGPT tells me Glances needs to monitor the host's network interfaces. Using host network mode enables it to see and track real-time traffic on the host's network interfaces without being isolated within a container-specific virtual network. And if you are running other services on the host machine that Glances needs to monitor, such as local databases or web servers, host mode allows the container to interact with them directly without needing complex network configurations or port mappings.

I installed without `network_mode: "host"` and Homepage can't see Glances (or more likely, Glances can't see the host so it has nothing to report).

glances/docker-compose.yaml
EnigmaCurry: In the case of homepage we made it optional, because not everything in homepage requires it. So I would prefer that if there are still usable features of glances that work without it.

mcmikemn: If we decide to keep Glances in d.ry, should we just eliminate the docker socker access?

glances/docker-compose.yaml
EnigmaCurry: ok maybe I need to relent on both of these issues, and just let it do its thing. We need a huge warning on the top of the README that this is essentially a system root level container.

glances/docker-compose.instance.yaml
EnigmaCurry: This is the behaviour of `network_mode: host` same as traefik. In effect, Docker has no clue what your container is doing network wise, because its taken over the host network.

EnigmaCurry: so this means that probably 61209 is probably already open, but your external firewall probably blocks it.

glances/docker-compose.yaml
EnigmaCurry:
```
## Warning

This container uses the docker flags `privileged: true` and
`network_mode: host`. This gives Glances unlimited access to your host
system. You should not install this unless you completely trust this
service.
```

EnigmaCurry:
https://github.com/nicolargo/glances/tree/24c87253595f9d3168afc26e82cfa464fc7523b9?tab=readme-ov-file#docker-the-cloudy-way

This example does use network mode host, but not privileged. Is privileged necessary?

EnigmaCurry:
```
## Warning

This container uses the docker flag `network_mode: host`. This gives Glances unlimited access to your host
network. You should not install this unless you completely trust this service. To enforce the use of traefik, your external firewall should block TCP ports 61208 and 61209.
```

EnigmaCurry:
With the addition of the README warning above, I am happy with this as-is.

EnigmaCurry (dismissed):

glances/.env-dist
mcmikemn: Fixed in [e86bce8](https://github.com/EnigmaCurry/d.rymcg.tech/pull/284/commits/e86bce8c4914bb96cc0a5d4f9d7427fc8c3c1ac5)

glances/docker-compose.yaml
mcmikemn: Fixed in [e86bce8](https://github.com/EnigmaCurry/d.rymcg.tech/pull/284/commits/e86bce8c4914bb96cc0a5d4f9d7427fc8c3c1ac5)

glances/docker-compose.yaml
mcmikemn: Fixed in [e86bce8](https://github.com/EnigmaCurry/d.rymcg.tech/pull/284/commits/e86bce8c4914bb96cc0a5d4f9d7427fc8c3c1ac5)

mcmikemn:
I reduced priveleges, added alert, variablized bind mount to docker socket, updated readme.

You were already OK with this PR with only the alert added. Do you want to review my changes, @EnigmaCurry?

glances/docker-compose.instance.yaml
```diff
@@ -37,6 +38,13 @@ services:
               capabilities: [gpu]
     #@ end
     
+    volumes:
+      #@ if container_metrics:
+      - /var/run/docker.sock:/var/run/docker.sock:ro
+      #@ end
+      #! The following line causes Glances to display host's OS detail instead of container's
+      - /etc/os-release:/etc/os-release:ro
```
EnigmaCurry, line 46: Will you also test if you can do the timezone and localtime as shown in #285 and then you probably dont need the TZ env var at all.

EnigmaCurry (dismissed):

glances/docker-compose.instance.yaml
mcmikemn: Didn't work. We decided to approve this PR using $TZ.

EnigmaCurry (approved):

PR #286, opened 2024-09-25
Add bind-mount host / to Glances

mcmikemn:
I added the ability to bnd mount `/` on host so Glances could report on disk space available on the host. What do you think?

### DIFF
--- a/glances/.env-dist
+++ b/glances/.env-dist
@@ -54,6 +54,10 @@ GLANCES_GPU=false
 # manipulate containers, and potentially compromise the host system.
 GLANCES_CONTAINER_METRICS=false
 
+# Set to "true" to have the Glances container bind mount `/` on the host so
+# that Glances can report on disk space available on the host, or set to "false".
+GLANCES_HOST_HD=false
+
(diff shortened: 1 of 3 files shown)


EnigmaCurry (approved):

EnigmaCurry (changes requested):
Can we rename the var like GLANCES_MOUNT_HOST_FILESYSTEM ?

EnigmaCurry:
Please test with an additional mounted volume too, does it get reported as well?

mcmikemn:
> Can we rename the var like GLANCES_MOUNT_HOST_FILESYSTEM ?

Fixed in [95ba59e](https://github.com/EnigmaCurry/d.rymcg.tech/pull/286/commits/95ba59e755cee4f48934cc754a6835bfc96e12f2)

mcmikemn:
> Please test with an additional mounted volume too, does it get reported as well?

Yes, Glances can report on multiple disks and/or filesystem mounts.

EnigmaCurry (approved):

PR #303, opened 2024-10-03
Update traefik-forward-auth and build from source for multi-arch

### LINKED ISSUE #302: Update traefik-forward-auth and build for multi-arch
The upstream docker image (`thomseddon/traefik-forward-auth:2`) is broken on arm64 (it apparently was mislabeled and was only built for amd64 instead.)

The image is also very out of date, so it would be better all around to just compile it from source.

EnigmaCurry:
Fixes #302


EnigmaCurry:
@mcmikemn I have tested this on raspberry pi (arm64) can you review this and test on amd64?

mcmikemn:
This built and installed fine, @EnigmaCurry. I don't know how to verify that it built the amd64 version, but I can see part of the command in the build history in the shell, so I think it worked.
```
=> [traefik-forward-auth builder 6/6] RUN if [ "linux/amd64" = "linux/arm64" ]; then       export GOARCH=arm64;     elif [ "linux/amd64" = "linux/arm" ] && [ "" = "v7" ]; then       export GO  20.0s
```

EnigmaCurry:
If it ran and it worked, it works.

PR #336, opened 2024-10-12
MeloTTS

EnigmaCurry:
A text to speech engine to potentially replace Larynx


EnigmaCurry:
could not get this to build.

PR #356, opened 2024-11-03
Invidious: Bypass google's block

### LINKED ISSUE #355: Invidious needs update to bypass Google's latest blocking method
Invidious hadn't worked for me for a couple months, even with `make reinstall` (the image tag is "latest"). 

I found in [Invidious installation instructions](https://docs.invidious.io/installation/#docker-compose-method-production) that you now need to submit identifying tokens to Google.

If this works, it makes your instance/traffic more traceable - they mention they are working on an automation to change the tokens periodically, but this could probably be scripted into d.ry. Maybe instead of asking the user to create and enter tokens, they can be automatically created and used on eveery `make install`. Or maybe there's a way to automatically create new tokens every hour in the image and tell Invidious to use the new values?

mcmikemn:
This will fix #355 .

### DIFF
--- a/invidious/.env-dist
+++ b/invidious/.env-dist
@@ -36,3 +36,20 @@ INVIDIOUS_REALM=Invidious
 INVIDIOUS_IP_SOURCERANGE=0.0.0.0/0
 
 INVIDIOUS_HMAC_KEY=
+
+# In the ongoing battle to bypass Google's attempt to block non-Google services
+# from scraping public Youtube videos and information, the current method for
+# Invidious to work is to pass a Proof of Origin Token to Google (if your
+# public IP is blocked by Google). To generate po_token and visitor_data
+# identities for passing all verification checks on the YouTube side, run
+# `docker run quay.io/invidious/youtube-trusted-session-generator` on the same
+# public IP address as the one blocked by YouTube, and copy the `visitor_data`
+# and `po_token` values here. The `po_token` and `visitor_data` tokens will make
+# your entire Invidious session more easily traceable by YouTube because it
+# is tied to a unique identifier - more info here:
+# https://docs.invidious.io/installation/#docker-compose-method-production
+INVIDIOUS_VISITOR_DATA=
+INVIDIOUS_PO_TOKEN=
+
+# Set to `true` to enable sending usage staticts to Invidious.
+INVIDIOUS_STATISTICS_ENABLED=false
--- a/invidious/.env-dist
+++ b/invidious/.env-dist
@@ -42,11 +42,11 @@ INVIDIOUS_HMAC_KEY=
 # Invidious to work is to pass a Proof of Origin Token to Google (if your
 # public IP is blocked by Google). To generate po_token and visitor_data
 # identities for passing all verification checks on the YouTube side, run
-# `docker run quay.io/invidious/youtube-trusted-session-generator` on the same
-# public IP address as the one blocked by YouTube, and copy the `visitor_data`
-# and `po_token` values here. The `po_token` and `visitor_data` tokens will make
-# your entire Invidious session more easily traceable by YouTube because it
-# is tied to a unique identifier - more info here:
+# `docker run --rm quay.io/invidious/youtube-trusted-session-generator` on the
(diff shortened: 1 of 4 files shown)


mcmikemn:
I have not gotten this to work yet. 

You need to create the tokens for a computer on the public IP so I'm running the command on sentry, but perhaps my floating IP is tripping me up?

EnigmaCurry:
I think you need to do it for your public ip address of your (home) docker server. The sentry vpn is only used for routing traffic into and out of your services only, not for general internet access.

EnigmaCurry:
You could make it work through the sentry by modifying these env vars in the traefik config:

```
TRAEFIK_VPN_ALLOWED_IPS 	Which IP subnets are routable by the VPN 	10.13.16.0/24, 0.0.0.0 (all traffic)
TRAEFIK_VPN_CLIENT_PEER_ALLOWED_IPS 	The VPN client allowed routable IP addresses 	10.13.16.1/32
```

mcmikemn:
> I think you need to do it for your public ip address of your (home) docker server. The sentry vpn is only used for routing traffic into and out of your services, not for general internet access.

Oh. Thanks for that suggestion. It works!

I get this in the inv_sig_helper logs:
```
inv_sig_helper-1  | 2024-11-04T13:57:41.504872454Z [2024-11-04T13:57:41Z WARN  inv_sig_helper_rust::player] nsig function ending did not work: =\s*function(\([\w]+\)\{\s*var\s+[\w\s]+=[\w\.\s]+?\.call\s*\([\w\s$]+?,[\(\)\",\s]+\)[\S\s]*?\}\s*return [\w\.\s$]+?\.call\s*\([\w\s$]+?\s*,[\(\)\",\s]+\)\s*\}\s*;)
inv_sig_helper-1  | 2024-11-04T13:57:41.515902610Z [2024-11-04T13:57:41Z INFO  inv_sig_helper_rust::player] sig code: var ECa;var RL={Ys:function(a){a.reverse()},
inv_sig_helper-1  | 2024-11-04T13:57:41.515935425Z     N3:function(a,b){var c=a[0];a[0]=a[b%a.length];a[b%a.length]=c},
inv_sig_helper-1  | 2024-11-04T13:57:41.515952628Z     vz:function(a,b){a.splice(0,b)}};ECa=function(a){a=a.split("");RL.vz(a,1);RL.N3(a,45);RL.N3(a,52);RL.vz(a,3);RL.Ys(a,31);RL.vz(a,3);RL.Ys(a,30);return a.join("")}
```
But Invidious works and videos can play.

I'll push this in a few minutes.

mcmikemn:
> You could make it work through the sentry by modifying these env vars in the traefik config:
> 
> ```
> TRAEFIK_VPN_ALLOWED_IPS 	Which IP subnets are routable by the VPN 	10.13.16.0/24, 0.0.0.0 (all traffic)
> TRAEFIK_VPN_CLIENT_PEER_ALLOWED_IPS 	The VPN client allowed routable IP addresses 	10.13.16.1/32
> ```

What effect would this have on other apps running behind that Traefik? Or on security in general?

EnigmaCurry:
> What effect would this have on other apps running behind that Traefik? Or on security in general?

I haven't tested it, but the idea is that you configure wireguard to allow some extra routes through the vpn, you could do it two ways:

 * Add the youtube.com IP networks to that env var (Don't know where you get that list). Th effect would be that the only traffic going through the VPN would be Traefik and Youtube.
 * Set the range to 0.0.0.0/0 so that all traffic from your docker server goes through the VPN.  The effect would be that any OS update, or any curl command, or any request at all would go through the VPN.
 
 I imagine your home IP is fairly static, but if you were wanting to access this from multiple sites (mobile) - having the proof of origin token generated on the sentry might work better. Or it might work worse if google shuts if off (like if you had several clients using it at the same time)

EnigmaCurry (approved):

mcmikemn:
@EnigmaCurry, are you still OK to merge this PR?

EnigmaCurry:
yes I have not tested it, but happy to have this make it better.

PR #377, opened 2024-12-08
Add env var to accommodate telemetry API change.

mcmikemn:
Immich updated their telemetry API and while the Homepage widget has a new version to accommodate the change, the Immich docker install needs to include a new env var.


EnigmaCurry (commented):
Telemetry always sounds scary, so it should go into the .env file with some better description of what it does. Should it really be turned on by default? Does it send any info anywhere outside your network?

PR #379, opened 2024-12-23
Add speedtest-tracker.

mcmikemn:
Add Speedtest Tracker.


EnigmaCurry:
> The logs say "generating self-signed keys in /config/keys, you can replace these with your own keys if required". 

If it says that in the app log (not the traefik log) then it seems like the app is trying to do its own TLS. you just need to turn that off and let traefik be the only thing that does TLS.

speedtest-tracker/docker-compose.yaml
```diff
@@ -0,0 +1,32 @@
+services:
+  speedtest-tracker:
+    image: ${SPEEDTEST_TRACKER_IMAGE}
+    restart: unless-stopped
+    #ports:
+    #    - 8080:80
```
EnigmaCurry, line 6: these old example ports suggest that it was designed to listen on both ports, so is there a config to turn off the second port?

speedtest-tracker/docker-compose.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - DB_CONNECTION=sqlite
+      - ADMIN_NAME=${SPEEDTEST_TRACKER_ADMIN_NAME}
+      - ADMIN_EMAIL=${SPEEDTEST_TRACKER_ADMIN_EMAIL}
+      - ADMIN_PASSWORD=${SPEEDTEST_TRACKER_ADMIN_PASSWORD}
+      - DISPLAY_TIMEZONE=${SPEEDTEST_TRACKER_DISPLAY_TIMEZONE}
+      - APP_URL=${SPEEDTEST_TRACKER_TRAEFIK_HOST}
+      - ASSET_URL=${SPEEDTEST_TRACKER_TRAEFIK_HOST}
+    volumes:
+      - data:/config
+      - keys:/config/keys
```
EnigmaCurry, line 21: I'm guessing the keys mount is to hold the TLS key, which is not needed.

speedtest-tracker/traefik-key.pem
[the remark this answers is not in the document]
mcmikemn: Oh, these were tests and I forgot to .gitignore them. Oops. I'll remove them, and I guess reinstall traefik to get a new key. :(

speedtest-tracker/docker-compose.yaml
mcmikemn: I don't see an env var or config to disable nginx's TLS. If you don't create the keys volume, it still creates its own keys but I guess just stores them ephemerally in the container. I'm sure it's possible to build the image and have Dockerfile go in and change an nginx config, but at this point it's not worth it to me - too much effort for something i may never use or need.

mcmikemn:
Everything works now. Ready for review.

speedtest-tracker/Makefile
```diff
@@ -0,0 +1,50 @@
...
+	@${BIN}/reconfigure_ask ${ENV_FILE} SPEEDTEST_TRACKER_TRAEFIK_HOST "Enter the speedtest-tracker domain name" speedtest-tracker${INSTANCE_URL_SUFFIX}.${ROOT_DOMAIN}
+	@${BIN}/reconfigure ${ENV_FILE} SPEEDTEST_TRACKER_INSTANCE=$${instance:-default}
+	@${BIN}/reconfigure_auth ${ENV_FILE} SPEEDTEST_TRACKER
+	@echo
+	@ALLOW_BLANK=1 ${BIN}/reconfigure_ask ${ENV_FILE} SPEEDTEST_APP_NAME "Enter the application's name as it will appear on the dashboard and in notifications. Or leave this bank to use the default (\"Speedtest Tracker\")"
+	@echo
+	@${BIN}/reconfigure_ask ${ENV_FILE} SPEEDTEST_TRACKER_ADMIN_NAME "Enter the admin user's login name" 
+	@${BIN}/reconfigure_ask ${ENV_FILE} SPEEDTEST_TRACKER_ADMIN_EMAIL "Enter the admin user's email" 
+	@${BIN}/reconfigure_ask ${ENV_FILE} SPEEDTEST_TRACKER_ADMIN_PASSWORD "Enter the admin user's password" 
+	@echo
```
EnigmaCurry, line 16: The README says this password is temporary for initial setup purposes only. I recommend that the Makefile just uses a random pw from reconfigure_password and then echo it to the user and tell them its a temporary password. theres no point in the user coming up with a password if they need to change again anyway. Also the .env-dist file should be clear that its an initial password only. I would name it SPEEDTEST_TRACKER_ADMIN_INITIAL_PASSWORD.

EnigmaCurry (changes requested):

speedtest-tracker/Makefile
EnigmaCurry: README says ADMIN_NAME and ADMIN_EMAIL are also only initial settings (ignored on subsequent starts) so I would name them to reflect that and add that info into .env-dist as a comment.

mcmikemn: I know their docs say that, but it doesn't appear to be the case. I can log into a fresh install via the admin login I define in env vars, and I'm not prompted to change the password and the admin user is listed in the Users section. I can create *additiona*  users, but it appears that the initial admin account is *not* temporary. I'll test again to be sure.

EnigmaCurry: but if you change the admin password, it keeps it? or it continues to work with the one from .env?

mcmikemn: The initial admin user that gets created from ADMIN_NAME and ADMIN_EMAIL is not temporary - it is permanent. The password assigned to this initial admin user from ADMIN_PASSWORD is changeable in the UI. 

In the UI, this initial admin user can create additional regular users and admin users, all of which are editable and deletable.

Because the initial admin user is permanent (thought the initial password can be changed), do you still think it's best to assign it a random password and let the user change it in the UI if they want?

FYI: it won't let you create multiple users with the same email address.

speedtest-tracker/Makefile
```diff
@@ -0,0 +1,50 @@
...
+	@${BIN}/reconfigure_ask ${ENV_FILE} SPEEDTEST_SCHEDULE "Enter the cron expression used to run speedtests on a scheduled basis" 
+	@echo
+	@${BIN}/reconfigure_ask ${ENV_FILE} SPEEDTEST_PRUNE_RESULTS_OLDER_THAN "Enter the number of days to keep results for, or enter 0 (zero) to keep all results" 
+	@echo
+	@${BIN}/reconfigure ${ENV_FILE} SPEEDTEST_TRACKER_APP_KEY=$$(openssl rand -base64 32)
+	@${BIN}/reconfigure ${ENV_FILE} SPEEDTEST_TRACKER_APP_KEY=$$(key=$$(${BIN}/dotenv -f ${ENV_FILE} get SPEEDTEST_TRACKER_APP_KEY); [[ "$${key}" == base64:* ]] && echo "$${key}" || echo "base64:$${key}")
+	@echo
+	@${BIN}/reconfigure_ask ${ENV_FILE} SPEEDTEST_TRACKER_UID "Enter the host UID to map to the container UID" 
+	@${BIN}/reconfigure_ask ${ENV_FILE} SPEEDTEST_TRACKER_GID "Enter the host GID to map to the container GID"
```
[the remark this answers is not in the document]
mcmikemn: Removed UID/GID as env vars, it works fine without them. Fixed in [2bf0d6d](https://github.com/EnigmaCurry/d.rymcg.tech/pull/379/commits/2bf0d6d090cd0f00f9e9421cf8e754462fb88d53)

speedtest-tracker/Makefile
EnigmaCurry: The default assumption of d.ry apps is that if there is a variable in a .env file, that you can change that var at any time, and that when you run `make install` the app will reflect that change. Rule #1 of d.ry is that config is set in the .env file. 

This apps breaks this assumption with some of these vars, because you can only set them in .env once (before volume creation) and you can only  change them in the app from that point on.

That is why I wanted the names to reflect that they are initial and should have a comment that they must be changed in the app from that point on.

However, looking at the config docs again theres actually several more that it says this is true of. I guess I withdraw my request because theres too many variables like this to try and fix.

I change my request: Instead of renaming them, I would just put a fat warning above these vars in the .env-dist that says that they can't necessarily be changed from the .env file.

mcmikemn: > However, looking at the config docs again theres actually several more that it says this is true of. 

Do you mean there are several d.ry apps that include env vars that must be edited in the app, or only that there are several env vars in speedtest-tracker like this?

EnigmaCurry: I meant in speedtrack there is more vars that say they are initial only: THRESHOLD_ENABLED, THRESHOLD_DOWNLOAD, etc. So I feel my first suggestion is fighting this, so I rescind it.

mcmikemn: Ah, I see. Well I think rule #1 is a good one, so despite there being multiple env vars in speedtest that the docs claim are initial only but are infact editable in the app, I think we should follow rule #1 and if an env var is *not* editable by changing the .env file and running `make install`, then we should *not* include it in the .env file.

So we'd *not* let the user define the admin username, email, or password - they'd log in with the app's defaults and then could change them in the UI. But as far as I can tell, all the other env vars I'm using in `docker-compose.yaml` are *not* configurable in the app and *are* editable with `make install`, so I'll leave them.

What do you think?

mcmikemn: In [3288100](https://github.com/EnigmaCurry/d.rymcg.tech/pull/379/commits/3288100e40f8f6d095f9c9daf95ac2857964da3f) I removed the 3 "initial admin" env vars. The app has default values built in, which I added to the README. Users can change them in the app.

EnigmaCurry: I think it would be more secure if the password was randomized via the Makefile rather than it being static. Some bot might register it before you do.

mcmikemn: Fixed in [9bb194c](https://github.com/EnigmaCurry/d.rymcg.tech/pull/379/commits/9bb194c633e585a95e7436485ac1968b3f2700cd)

EnigmaCurry: good :+1:

EnigmaCurry (approved):

EnigmaCurry:
I love that these just happen to be listed alphabetically together

[Smokeping](https://github.com/EnigmaCurry/d.rymcg.tech/blob/master/smokeping#readme) - a network latency measurement tool
[Speedtest Tracker](https://github.com/EnigmaCurry/d.rymcg.tech/blob/master/speedtest-tracker#readme) - a network performance monitor

Issue #393, opened 2025-02-13
Add Frigate.


mcmikemn:
I'd like to use Frigate to manage video cams for Home Assistant.

PR #394, opened 2025-02-14
Frigate

mcmikemn:
Implements #393

### DIFF
--- a/frigate/Makefile
+++ b/frigate/Makefile
@@ -3,16 +3,40 @@ include ${ROOT_DIR}/_scripts/Makefile.projects
 include ${ROOT_DIR}/_scripts/Makefile.instance
 
 .PHONY: config-hook
+.ONESHELL:	
 config-hook:
-#### This interactive configuration wizard creates the .env_{DOCKER_CONTEXT}_{INSTANCE} config file using .env-dist as the template:
-#### reconfigure_ask asks the user a question to set the variable into the .env file, and with a provided default value.
-#### reconfigure sets the value of a variable in the .env file without asking.
-#### reconfigure_htpasswd will configure the HTTP Basic Authentication setting the var name and with a provided default value.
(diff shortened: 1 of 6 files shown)


mcmikemn:
@EnigmaCurry: this is working in that Frigate gets installed, the basics get configured, and you can log into it. I don't have any cameras to test video streaming so it's possible that I'll need to add something to `docker-compose.instance.yaml` for RTSP and WebRTC (https://docs.frigate.video/frigate/installation/#docker).

frigate/Makefile
```diff
@@ -0,0 +1,76 @@
...
+	@make --no-print-directory docker-compose-shell SERVICE=frigate
+
+.PHONY: show-admin-user # Show the automatically-generated admin login credentials
+show-admin-user:
+	@echo
+	@docker logs frigate-frigate-1 2>&1 | grep -F '***' || echo "Error: admin credentials were not found in logs."
+	@echo
+
+.PHONY: disable-tls # Disable TLS on Frigate's webserver
+disable-tls:
```
EnigmaCurry, line 67: Normally in d.ry we setup a "config" container to setup any non env var enabled config. This would set the config file before the service starts. Can we do that instead?

EnigmaCurry:
For the extra ports needed, theres a few options:

Add entrypoint through traefik. This has the benefit that you can bind to specific host ip address and apply an ip whitelist. But this has the overhead of passing large video frames through traefik.

Add the port mapping directly in docker compose. We dont normally do this, but it might be preferred for constant video ingest.(no traefik overhead) however you lose the ip filter so you would want to be sure to bind it to a specific host ip address so its not publicly accessible.

frigate/Makefile
mcmikemn: That is, of course, better way. Fixed in [10c27c6](https://github.com/EnigmaCurry/d.rymcg.tech/pull/394/commits/10c27c6f9ffee671c586ea0375ccb69372624c1a)

Issue #400, opened 2025-03-02
Add Sablier.


mcmikemn:
There are multiple d.ry apps that I rarely use - my current options are:
- leave them running all the time, using resources when not being used
- manually start them up when I do want to use them

The current options are absolutely do-able, but the situation would be improved if the containers would start up only when they're needed/requested and automatically shut down after a period of inactivity.

[Sablier](https://github.com/sablierapp/sablier) starts containers on demand and stops them after a period of inactivity.

EnigmaCurry:
I'm not opposed to adding sablier, but I have an alternate proposal in #402 

I'm not personally interested in Sablier, for a couple reasons:

 * I don't want a random web request to start my containers, I want it to be an admin action only.
 * I might want to do this for non-web services.

#402 is my alternative proposal for an LLM agent that satisfies that same sort of dynamic deployment I'm after.

PR #401, opened 2025-03-02
Add Sablier Traefik plugin.

mcmikemn:
Sablier may or may not provide beneficial functionality, and may or may not be a good fit for d.ry. I'm trying to add it in order to check it out. 

If we add it, this would implement #400

### DIFF
--- a/traefik/.env-dist
+++ b/traefik/.env-dist
@@ -12,7 +12,7 @@ TRAEFIK_CONFIG_VERBOSE=false
 ## Or use a sidecar service like a VPN client: TRAEFIK_NETWORK_MODE=service:wireguard-client
 TRAEFIK_NETWORK_MODE=host
 
-## Enable optional docker-compose profiles: default,error_pages,geoip_update,wireguard,wireguard-client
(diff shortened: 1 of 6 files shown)


mcmikemn:
This project turned out to be beyond my skillset. It requires:
- adding a Traefik plugin
- possibly needing to template a docker-compose.override.yaml file for traefik installation ([see here](https://github.com/EnigmaCurry/d.rymcg.tech/blob/49a7c3e9bec8eb6328af4e01f33e523c42357a8e/traefik/docker-compose.yaml#L198))
- possibly defining Docker labels for apps that use Sablier differently than d.ry currently does ([see here](https://sablierapp.dev/#/plugins/traefik?id=prerequisites))

I'm going to shelve this project. If you think it's worth looking into, @EnigmaCurry, I'd love to work on it with you (even if that means that I just watch you and ask questions). If you're not interested in it, we should eventually just close this draft PR and the corresponding [issue](https://github.com/EnigmaCurry/d.rymcg.tech/issues/400).

PR #415, opened 2025-04-15
Implement Layer 4 UDP routes

### LINKED ISSUE #414: Add layer 4 udp  routing
~Jitsi-meet requires port 10000 udp for video/audio, so it must be opened on the firewall and a layer 4 route must be created in the sentry Traefik.~

Implement UDP for layer 4 routes

EnigmaCurry:
Fixes #414 

This implements udp routing but does not fix jitsi meet (it must have some other issue)


EnigmaCurry:
@mcmikemn I have refactored this to handle both stock and custom entrypoints and layer 4 routes for tcp and udp.

I have added the iperf bandwidth testing program which supports both tcp and udp. I have tested that it works.

I would like you to test this branch by installed traefik and see if your existing services are all still working. You can test iperf if you want, but I have already tested that works, so I mostly just want to confirm I didnt break your existing stuff.

mcmikemn (approved):

PR #425, opened 2025-04-23
Add pictrs config to auto-prune.

mcmikemn:
Add configurations to pictrs service telling it to prune old content.

### DIFF
--- a/lemmy/.env-dist
+++ b/lemmy/.env-dist
@@ -57,7 +57,7 @@ LEMMY_PRIVATE=false
 LEMMY_OAUTH2=false
 # In addition to Oauth2 authentication, you can configure basic authorization
 # by entering which authorization group can log into your app. You create
-# groups of email addresses in the `traefik` folder by running `make groups`. 
+# groups of email addresses in the `traefik` folder by running `make groups`.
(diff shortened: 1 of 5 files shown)


EnigmaCurry (approved):
It looks good. One nit is you should probably use a separate volume for the pictrs config, and mount both to the config container. But its not terribly important I imagine.

mcmikemn:
> It looks good. One nit is you should probably use a separate volume for the pictrs config, and mount both to the config container. But its not terribly important I imagine.

I thought of doing that but ended up using a single config container for both services. The lemmy config volume isn't needed by pictrs at all. If you'd prefer, I'll switch it to a separate config container.

EnigmaCurry:
Im not suggesting a separate config container, just two config volumes attached to the config container. lemmy-config and pictrs-config.

mcmikemn:
> Im not suggesting a separate config container, just two config volumes attached to the config container. lemmy-config and pictrs-config.

Is [this](https://github.com/EnigmaCurry/d.rymcg.tech/pull/425/commits/21cf4cafb6b187e1e5a2dc9ac3eb9a6263db3373) what you meant?

EnigmaCurry:
Yep

EnigmaCurry (approved):

PR #431, opened 2025-05-21
Rclone Docker Volume Plugin and File Browser file manager

EnigmaCurry:
This adds the [Rclone Docker Volume Plugin](https://rclone.org/docker/#introduction) and the [File Browser](https://github.com/filebrowser/filebrowser?tab=readme-ov-file) file manager.

See [RCLONE.md](https://github.com/EnigmaCurry/d.rymcg.tech/blob/5d8f263ecb6763af414e3521be22d553e4a67ff8/RCLONE.md)


EnigmaCurry:
@mcmikemn I added the ability to mount an S3 bucket into the filebrowser container. With this change you can configure an S3 backed volume for _any_ container, without the app needing to support S3 directly.

mcmikemn (approved):

_scripts/s3_volume_create
```diff
@@ -60,7 +60,10 @@ if ! docker plugin inspect "${DRIVER}" 2>&1 >/dev/null; then
     esac
     docker plugin install "rclone/docker-volume-rclone:${PLUGIN_ARCH}" \
            args="-v" --alias rclone --grant-all-permissions
-    docker plugin enable rclone
+    # Enabling the plugin after installing it without `--disable` is
+    # not necessary, but I'm leaving it here just in case it's needed
+    # for idempotency
+    docker plugin enable rclone 2>/dev/null
```
EnigmaCurry, line 66: hmm, when I installed it was disabled by default thats why I added this.

mcmikemn: When I installed, it was already enabled. My workstation is debian.

I left it in, just redirected stderr to null, so it should still work for both situations.

filebrowser/docker-compose.yaml
```diff
@@ -41,6 +41,6 @@ services:
     labels: []
 
 volumes:
-  srv:
+  # srv:
```
EnigmaCurry, line 44: So it should be removed?

_scripts/s3_volume_create
EnigmaCurry: ok but not sure redirecting stderr is a good idea

filebrowser/docker-compose.yaml
mcmikemn: Yes. Fixed in [7a2088b](https://github.com/EnigmaCurry/d.rymcg.tech/pull/431/commits/7a2088bcd3d03ecd5672504b5dc9a325fa9c5a3a)

_scripts/s3_volume_create
mcmikemn: Good thinking. Fixed in [10e8a4f](https://github.com/EnigmaCurry/d.rymcg.tech/pull/431/commits/10e8a4f5803b8fdc20fdecb4440088d6216cc18a)

mcmikemn (approved):

PR #438, opened 2025-05-28
Variablize service versions

mcmikemn:
It's possible that we'll be removing jitsi from d.ry, but as longs as it's still here, I had this change from when we were last trying to make it work: variablize the service versions.


mcmikemn:
> I think this duplicates the stable- prefix?

Oops. Fixed in [546455d](https://github.com/EnigmaCurry/d.rymcg.tech/pull/438/commits/546455d50e952751cd3c28e33d39109805a9dcd3)

EnigmaCurry (approved):

Issue #440, opened 2025-06-04
Add plugins to redmine


mcmikemn:
Add plugin functionality to Redmine, include Gantt plugin. (https://www.easyredmine.com/redmine-gantt-plugin).

ChatGPT suggestion:

Folder Structure
```
redmine-docker/
├── docker-compose.yml
├── Dockerfile
└── plugins/
    └── redmine_example_plugin/   # Cloned plugin repo
```
Dockerfile
```
FROM redmine:6.0

# Copy plugins into image
COPY plugins /usr/src/redmine/plugins

# Install dependencies and migrate
WORKDIR /usr/src/redmine
RUN bundle install && \
    bundle exec rake redmine:plugins:migrate RAILS_ENV=production
```

docker-compose.yml
```
services:
  redmine:
    build: .
    ports:
      - "3000:3000"
    volumes:
      - redmine_data:/usr/src/redmine/files
    environment:
      REDMINE_DB_MYSQL: db
      REDMINE_DB_DATABASE: redmine
      REDMINE_DB_USERNAME: redmine
      REDMINE_DB_PASSWORD: secret
      REDMINE_DB_ENCODING: utf8

  db:
    image: mysql:5.7
    environment:
      MYSQL_ROOT_PASSWORD: rootpass
      MYSQL_DATABASE: redmine
      MYSQL_USER: redmine
      MYSQL_PASSWORD: secret
    volumes:
      - mysql_data:/var/lib/mysql

volumes:
  redmine_data:
  mysql_data:
```

PR #441, opened 2025-06-05
Add plugin functionality.

mcmikemn:
Implements #440


mcmikemn:
I included the easy_gantt plugin as an example but I can see not wanting to include it in d.ry. I'm good either way.

mcmikemn:
> I think it would be better if they were loaded in the Dockerfile from a remote repository, they shouldn't be here in this repo.

The original intent was to have the install-plugin process for a given context-instance copy the `plugins-example` dir. I ended up not doing it that way, but left the easy_gantt plugin in the example dir so users wouldn't have to find/download it. But I agree with you, so I completely removed the `plugins-example` dir in [c025743](https://github.com/EnigmaCurry/d.rymcg.tech/pull/441/commits/c0257437b6a6850b7ede0dc740dd65d863f5d319).

EnigmaCurry (approved):

PR #445, opened 2025-07-07
Bumps

mcmikemn:


EnigmaCurry:
Does the final diff have all your changes? I don't see vaultwarden, peertube, actual, or homepage?

EnigmaCurry:
oh I see those are old commits, so this is just immich and audiobookshelf?

EnigmaCurry (approved):

mcmikemn:
I'm not sure how I did that, but yes: this PR is only the latest updates of audiobookshelf and immich.

PR #446, opened 2025-07-11
Add Open webui

mcmikemn:
Add Open-webui. Tested with rocm, cuda, and cpu.

### DIFF
--- a/_scripts/reconfigure_compose_profiles_select
+++ b/_scripts/reconfigure_compose_profiles_select
@@ -1,7 +1,8 @@
 #!/bin/bash
 
-## reconfigure_compose_profiles_select ${ENV_FILE} profile="description" ...
-## wizard select some profiles and then call reconfigure_compose_profiles
+## reconfigure_compose_profiles_select ${ENV_FILE} mode profile="description" ...
+## wizard select some profiles (select) or choose a profile (choose)
+## and then call reconfigure_compose_profiles
 
 BIN=$(dirname ${BASH_SOURCE})
 source ${BIN}/funcs.sh
@@ -11,6 +12,9 @@ ENV_FILE=${1}
 shift
 ## Make new .env if it doesn't exist:
 test -f ${ENV_FILE} || cp .env-dist ${ENV_FILE}
+
+MODE=${1:-select}
+shift
 
 readarray -t ALL_PROFILES < <(cat docker-compose.yaml | yaml_to_json | jq -r '.services[].profiles | values[]' | sort -u)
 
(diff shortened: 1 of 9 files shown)


_scripts/reconfigure_compose_profiles_select
```diff
@@ -1,7 +1,8 @@
 #!/bin/bash
```
EnigmaCurry, line 1: I guess the script should be renamed reconfigure_compose_profiles now that there are two options for choose or select.

open-webui/docker-compose.instance.yaml
```diff
@@ -0,0 +1,88 @@
...
+  ollama-cuda:
+    volumes:
+      - (@= models_host_path @):/root/.ollama/models
+  #@ end
+  #@ if compose_profile == "cpu":
+  ollama-cpu:
+    volumes:
+      - (@= models_host_path @):/root/.ollama/models
+  #@ end
+  #@ end
```
EnigmaCurry, line 44: I think the templated conditional is unnecessary since you are using compose profiles. just put them all statically should work and only the profile that is selected will be configured.

open-webui/docker-compose.yaml
```diff
@@ -0,0 +1,62 @@
...
+      - /dev/dri
+    security_opt:
+      - seccomp=unconfined
+    group_add:
+      - video
+      - render
+    environment:
+      # HSA_OVERRIDE_GFX_VERSION should probably be variablized
+      - HSA_OVERRIDE_GFX_VERSION=11.0.0
+    volumes:
```
EnigmaCurry, line 42: something needs to be installed in the host (VM)? Can you please document that?

_scripts/reconfigure_compose_profiles_select
mcmikemn: Updated in [ce83ac7](https://github.com/EnigmaCurry/d.rymcg.tech/pull/446/commits/ce83ac7442c5f9a9ed6cfd3fc64790fa99cda198)

open-webui/docker-compose.instance.yaml
mcmikemn: Actually, the volume needs to be templated anyway because if the user doesn't want to bind mount a host path (i.e., if OPENWEBUI_MODELS_HOST_PATH has no value), then we don't want to declare that volume.

open-webui/docker-compose.yaml
```diff
@@ -0,0 +1,62 @@
...
+      - /opt/rocm:/opt/rocm
```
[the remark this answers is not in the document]
mcmikemn: I removed it and everything works. It's not mentioned in the [very limited] ollama:rocm docs in hub.docker.com or on ollama's gpt repo. I can only imagine it must have been a hallucination earlier in the process.

Removed in [22fc0fe](https://github.com/EnigmaCurry/d.rymcg.tech/pull/446/commits/22fc0fe36a3253a2bb5d096554b108776e980ca7)

PR #484, opened 2025-09-04
Adds Comfyui

### LINKED ISSUE #483: Add ComfyUI
InvokeAI is a good AI image generator, and ComfyUI does do image generation, but ComfyUI also generates video, audio, and 3D models. And while the image generation is kind of redundant in d.ry, the interfaces are vastly different: InvokeAI is more user friendly (believe it or not!) and ComfyUI is node-based (i.e., you need to be more tech-savvy).

mcmikemn:
Resolves #483

### DIFF
--- a/comfyui/model-manager.sh
+++ b/comfyui/model-manager.sh
@@ -54,12 +54,12 @@ while true; do
     model_type=$("${BIN}/script-wizard" choose "What type of model are you installing?" ${choices} --default "checkpoints")
     echo ""
     model_url=$("${BIN}/ask_echo" "What is the download URL of the model?" "")
+    echo ""
     if [[ "${model_url}" =~ .*civitai\.com.* ]]; then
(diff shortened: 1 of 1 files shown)


mcmikemn:
- This works for ROCm
- I need to test CPU-only more
- I bet tweaks will be needed before it'll work with Nvidia GPUs - can you please test with Nvidia when you get a chance, @EnigmaCurry?
- I'm not sure we have the means to test Intel GPUs. I added it because COmfyUI supports them. We can either leave it in and if anyone ever tries to run with Intel GPU and has a problem we can update it; or we can remove Intel.

EnigmaCurry:
works with cuda. I'm fine with adding the untested hardware modes.

EnigmaCurry:
Many of the templates require multiple model files that the operator must download separately via `make add-model`. An improvement would be to enter all of the models to download at once at the beginning in a loop.

PR #518, opened 2025-10-05
Fix ComfyUI update mechanism

### LINKED ISSUE #517: Can't update ComfyUI from UI
ComfyUI Manager has a feature to update ComfyUI from within the UI itself (and to switch between versions), but it throws an error.

mcmikemn:
ComfyUI Manager has a feature to update ComfyUI from within the UI itself (and to switch between versions), but it throws an error.

The problem is that Dockerfile clones the specific tag for the version the user wants to install - it doesn't clone the master branch, which ComfyUI Manager's "Update ComfyUI" function requires.

Fixes #517

### DIFF
--- a/comfyui/Dockerfile
+++ b/comfyui/Dockerfile
@@ -53,5 +53,6 @@ ARG HSA_OVERRIDE_GFX_VERSION
 ENV HSA_OVERRIDE_GFX_VERSION=${HSA_OVERRIDE_GFX_VERSION}
 ARG LOG_LEVEL
 ENV LOG_LEVEL=${LOG_LEVEL}
+VOLUME /ComfyUI
 ENTRYPOINT ["/entrypoint.sh"]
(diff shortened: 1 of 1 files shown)


mcmikemn:
> Does that mean that you need to remove COMFY_VERSION form .env and hard code it to master?

No. You can still choose the version to install, but now Dockerfile will clone the specified version *and* the master branch, and will install the specified version. The master branch is cloned so that ComfyUI Manager can access it for the "Update ComfyUI" and "Switch ComfyUI [version]" features. I didn't look at their code to see why they need master to change versions.

mcmikemn:
> Not having tried the feature, I would expect any in-place update to require a VOLUME listed in the Dockerfile.

Can you explain that?

There is a named volume defined in `docker-compose.yaml`, so my thinking is that once installed, if you update it from within the UI, the updated files are all persistent on the named volume.

PR #532, opened 2025-10-17
Create separate ollama service

### LINKED ISSUE #529: Create separate Ollama
* Remove ability of open-webui to expose it's internal Ollama service.
  * Add bool flag OPENWEBUI_OLLAMA_ENABLE=true to enable the internal Ollama service.
  * Create a completely separate Ollama project for the use-case of exposing ollama.

mcmikemn:
Also reconfigure ollama service in the Open-WebUI stack to be enablabled/disabled, and to not be exposed.

Resolves #529


EnigmaCurry:
I added an optional API token to the exposed ollama.

EnigmaCurry (dismissed):
other than that, LGTM

PR #553, opened 2025-10-30
Refactor wireguard container router for qBittorrent, Firefox, and Aria2

### LINKED ISSUE #554: Wireguard container does not have a robust kill switch
Although the default route it sets up does go through the VPN, you can still leak packets through your native connection if you are resourceful enough:

```
## This will bypass the VPN completely because the eth0 device is still active in the container:
curl -4 --interface eth0 https://ifconfig.me
```

I doubt there is a pure docker approach that can prevent this. However, it should be possible to create a firewall rule on the host that prevents the default container interface from reaching the general internet, by dropping any packets not destined for its wireguard peer.

EnigmaCurry:
This gets rid of the `network_mode` approach of attaching to the wireguard namespace. Instead, wireguard now acts as a router, and containers that wish to route through the VPN must change their default route to do so. This requires NET_ADMIN capability, but we can quickly drop it in the entrypoint after creating the routing table changes.

Fixes #554


mcmikemn (approved):
This tested successfully with firefox and qbittorrent.

PR #583, opened 2026-01-15
Adds custom.css to Open-WebUI

### LINKED ISSUE #582: [FEATURE] would like a way to customize css for Open-WebUI
I find parts of the UI in Open-WebUI kind of ugly, and the admin panel hard to read. I'd like the ability to add a theme or skin to Open-WebUI - they don't support that, but you can override their CSS.

mcmikemn:
`make config` asks user if they want to add custom.css, copies it to openwebui container if they do (using install-hook target).

Resolves #582

### DIFF
--- a/open-webui/README.md
+++ b/open-webui/README.md
@@ -34,9 +34,10 @@ become the admin automatically.
 Optionally, you can create your own CSS stylesheet to customize
 Open-WebUI's appearance (remember to append the `!important`
 declaration to ensure your custom styles take precedence over
-Open-WebUI's internal style sheets). Be sure to save your custom CSS
-file in this project's directory before running `make install`, and
-name the filet `custom.css`.
+Open-WebUI's internal style sheets). Save your custom CSS file in this
+project's directory before running `make install`, and name the file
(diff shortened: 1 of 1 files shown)


open-webui/Makefile
```diff
@@ -60,6 +68,13 @@ shell:
 
 .PHONY: install-hook
 install-hook:
+	@if [[ "$$(${BIN}/dotenv -f ${ENV_FILE} get OPENWEBUI_CUSTOM_CSS)" == "true" ]]; then \
+		if [[ -f custom.css ]]; then \
+			docker cp custom.css open-webui-openwebui-1:/app/build/static/custom.css; \
+		else \
+			echo -e "\nNo \`custom.css\` file found in \`$$(pwd)/\`."; \
+		fi \
+	fi
```
EnigmaCurry: Build CONTAINER_NAME from INSTANCE 

Example: https://github.com/EnigmaCurry/d.rymcg.tech/blob/91e5518b66e581ed72c4681760711f82395360a9/redmine/Makefile#L42-L47

PR #585, opened 2026-01-15
Restores the matrix error-pages theme.

### LINKED ISSUE #584: Restore matrix theme in error-pages
It was removed upstream, but I have restored it in my fork and I've test it works.

I asked this question in the upstream discussion forum: 

https://github.com/tarampampam/error-pages/discussions/382

EnigmaCurry:
Fixes #584

### DIFF
--- a/traefik/docker-compose.yaml
+++ b/traefik/docker-compose.yaml
@@ -275,7 +275,6 @@ services:
       traefik.enable: true
       # use as "fallback" for any NON-registered services (with priority below normal)
       traefik.http.routers.error-pages-router.rule: HostRegexp(`.+`)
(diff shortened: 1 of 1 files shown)


EnigmaCurry:
https://doc.traefik.io/traefik/reference/routing-configuration/http/routing/rules-and-priority/#rulesyntax

The warning is a deprecation notice only, so I don't think that's it, but I removed the ruleSyntax line anyway.

Is the domain that shows the traefik default cert for an error page the same domain that is working for a different page? Sounds like the config for the domain is just messed up in totality, not just for error pages?

mcmikemn:
The problem was me, not the code. This PR LTGM.

